PR #18, opened 2022-07-25
Pdf to Markdown

mariobodemann:
Hey,

to increase the readability and accessibility of our documentation, I'm suggesting to favor open file formats like Markdown over closed formats.

Why? This helps in collaboration: If something is different then we want it in PDFs, we need to have the original file, and cannot easily make changes. And once those changes are made, no body understands who made those.

I suggest using markdown, and therefore started with a simple sample.


Kevsy:
Hi Mario, +1. In fact this is [closed issue](https://github.com/camaraproject/WorkingGroups/issues/36) for the Commonalities working group -  but as you say the rule needs to be applied throughout CAMARA.  We can use Mermaid syntax for sequence diagrams, use cases etc. within markdown as required.

kevin8xu (commented):
Agreed on using open format for documents whenever possible.

Kevsy:
closing as this is an existing issue (rather than PR) and we've put it into action with recent contributions.

PR #24, opened 2022-08-04
Uni api

AureMG:


maheshc01:
Adding few Pros for 5gff discovery API Spec to provide better understanding to everyone looking at the Analysis.

1. 5GFF Discovery API spans the compute and network continuum to offer a singular discovery that developer can use to discover a suitable MEC that not just answers the question of what is the closest MEC, but strives to answer the question, What is the closest MEC that meets my compute Requirement. It would be a poor Developer experience to discover the closest MEC and then realizing all of the MECs don’t meet the compute requirements. GPU Service not present in the MEC, which is needed by the workload.

2. 5GFF Discovery APIs are Intent based API, that are completely ACCESS agnostic, keeping true to the M in MEC (Multi-Access Edge Compute). The API is generic enough, it can be implemented not only for 4G(3GPP doesn’t cover 4G) and 5G, it can also be implemented to discover over the Fixed networks (which would be equally have large portion of discovery use cases). It would be impossible for a developer to use different Telco APIs for different ACCESS types.

3. 5GFF APIs are intent based APIs, that hides the complexity of the network and doesn’t require a developer to have deep understanding of how a mobile network operates to use the API.
 

Opinion on how we could move forward.
Rather than not looking at UNI spec as part of this work stream, we can reduce the fragmentation by looking at opportunities to extend features and use cases supported by UNI spec and abstract them as part of new revision of Edge Discovery API.

Kevsy:
Just to note that **UNI** is defined as the ’User Network Interface’ by the GSMA Operator Platform architecture. UNI is a middleware not required with the 5GFF API, but required for the GSMA approach.

From the GSMA Operator Platform documentation:
**User Client:** Functionality that manages on the user’s side the interaction with the OP. The User Client represents an endpoint of the UNI and is a component of the User Equipment. NOTE: Different implementations are possible, for example, OS component, separate application software component, software library, SDK toolkit and so on.

**UNI (User-Network Interface):** Enables the User Client (UC) hosted in the user equipment to communicate with the OP. 

**What this means:**  for the scenario of a  smartphone app calling the operator API there is  difference:

5GFF:
UE-hosted app calls operator API over HTTPS (no User Client required)

GSMA-compliant API:
UE-hosted app calls the GSMA User Client, which brokers the API call over the UNI to the operator API.

PR #39, opened 2022-10-11
Traffic Influence API proposal

FabrizioMoggio:
TI API proposal to be merged on the main branch

### DIFF
--- a/api_proposal/APIproposal_Traffic Influence_TIM.md
+++ b/api_proposal/APIproposal_Traffic Influence_TIM.md
@@ -1,0 +1,13 @@
+| **Field** | Description | 
+| ---- | ----- |
+| API name | Traffic Influence |
+| API owner | TIM, TEF |
+| API summary | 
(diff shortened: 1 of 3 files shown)


PrashantGDT:
How this API is different from QoD APIs or Network Slicing APIs in general where user can request for a different session?

Regards
Prashant Goswami

FabrizioMoggio:
> 

QoD, among other features, could be used to ask for a specific latency and this request can lead the Operator Platform  to invoke different network APIs to guarantee that latency. A possible solution, among others, is to invoke the 3GPP Traffic Influence API. The CAMARA TI API is stead a simplification, an abstraction, for the developer to directly and specifically request a special routing and maybe a specific QoS for a specific set of users. QoD and TI are supporting indeed different use cases both related to some quality aspects. 

Network Slicing is a more complex and wider solution that could lead to the deployment of UPFs at the edge as part of a specific slice, this could be an enabler for the CAMARA TI API but there is no overlapping

PR #42, opened 2022-11-07
Update edge_terminology.md

gunjald:
Added terminologies from Operator Platform Telco Edge Requirements Version 3.0


Kevsy:
Merged for discussion on sub-group call. It needs to be reordered alphabetically, but we can do that following discussion.

PR #46, opened 2022-11-24
Traffic Influence stable YAML and API Documentation

FabrizioMoggio:
Stable YAML, coherent with the Commonalities Security specification and API Documentation, coherent with the Commonalities Documentation specification


Kevsy (changes requested):
Line 126, 177: HTTP 201 is Created, not resource edited. 200 OK or 204 No Content are appropriate

Kevsy (approved):

PR #51, opened 2023-01-10
Create describing and harmonising the Edge APIs

Kevsy:


Kevsy:
Draft only, deleting this PR

PR #52, opened 2023-01-10
Create describing and harmonising the Edge APIs

Kevsy:
First draft for initial review. 

Note I have called the third API 'GSMA and MobiledgeXR' API for now but please rename it if it has another name

TODO:
I need to add the flows for each API


ThomasEdgeXR (approved):

gunjald (approved):
Approved

PR #55, opened 2023-01-18
Update and rename describing and harmonising the Edge APIs to describ…

Kevsy:
…ing and harmonising the Edge APIs.md

Included the EdgeXR and Cap Gemini APIs that are now in main


gunjald (approved):
Good information to proceed with.

PR #64, opened 2023-02-21
Edge cloud developer journey proposal

maheshc01:
listed recommendations of core modules and respective functionality that should be available for developers as part of this edge cloud api family.


gunjald (dismissed):
Initial version seems fine

Kevsy (changes requested):
'common features' is a column, so each feature appears in a cell of a row, which makes it look like each only applies to the row it appears in. But I think the intention is to list features common to all - please can you confirm, and if so re-format so that is clear?

PR #65, opened 2023-02-21
Update describing and harmonising the Edge APIs.md

gunjald:
The changes submitted describes the developer or application provider intents enabled by the OPAG/Capgemini proposed APIs on NBI


Kevsy (changes requested):
Hi Deepak, as per email please can you reformat the PR so that the new intents and GMEC label are included in the existing table?

Kevsy (approved):

PR #68, opened 2023-02-22
TI API v0.8.2

FabrizioMoggio:
TI API YAML and Documentation update.

API Changelog:
-POST and trafficInfluenceID: it has been better documented how to use it
-State: the different states have been described
-TIResource_Id and trafficInfluenceID: currently just trafficInfluenceID is used in the YAML
-FlowInfo/flowDescriptions: it is not used number anymore to identify the protocol but a text (e.g. TCP or UDP).


gunjald (approved):
Seems good.

PR #76, opened 2023-03-15
Corrected version to 0.8.0 as agreed in Feb call

Kevsy:


gunjald (approved):
version change looks fine

PR #100, opened 2023-05-16
Draft models for GSMA/EdgeXR and 5GFF

Kevsy:


Kevsy:
Apologies for pptx format - once discussed I can move it to Markdown

gunjald (commented):
In my opinion, the GSMA Model assumes the UNI interface to be lightweight function to assist User Clients(UCs) to discover edge applications, may request dynamic instantiation of edge apps, may provide indicators to QoS being experienced etc.

From this perspective I think in the slide #2, the UNI interface "All functional intents" are actually the Discovery Intent while the Developer to Operator interface i.e., GSMA NBI is about the Apps and artefacts Management and Orchestration intents, monitoring intents, developer management events, edge discovery intents etc which together we may call "All functional intents"

Kevsy:
Thanks @gunjald , agreed - I will make the change

PR #101, opened 2023-05-17
Use Case for User Mobility

FabrizioMoggio:
This is the use case for release 2 of the Traffic Influence API


gunjald (approved):
Seems fine

PR #110, opened 2023-07-04
Traffic Influence Release 0.9.1 to Dev Repo

FabrizioMoggio:
Beta candidate release


FabrizioMoggio:
this is the TI API Beta version, according to the discussion in Readiness Checklist step 6 update proposal #249 (https://github.com/camaraproject/WorkingGroups/issues/249)

gunjald (changes requested):
1. There is a small spelling mistake i.e., "ir" that may be changed to "is" in following statement "Identifier for the Traffic Influence resource. This parameter ir returned by the API and must be used to update it"
2. Should there be some guidance on on using zone ID vs Region ID? How a developer can decide which one to use when? Any suggestions?

FabrizioMoggio:
@gunjald 
1. thank you, I'm going to fix it
2. Zone ID and Region ID are defined by the Telco Operator who is sharing the Edge resources. My idea is that they are explained in the Marketplace portal where the TI API is sold. In the Portal the Developer can find a mapping among zones, regions and actual geographical locations. Maybe we can add this concept in the API Documentation. Anyway **the same explanation should be also included in the other Edge Cloud APIs if zones and regions are used**

gunjald:
The PUT operations typically affects the mutable attributes of a resource. Generally absence of such attributes in the PUT request is assumed by server implementation to remove or unset the given attribute. 

In TI API i think we should also capture what are the mutable attributes and what are immutable attributes and that would also help server implementation to check and update only mutable attributes of the TI API resource. Or we can even think of using PATCH operation to selectively update the mutable attributes.

FabrizioMoggio:
@gunjald it is a good suggestion. I would prefer to start reasoning on possible improvements still using the PUT method rather than introducing PATCH, if possible, to reduce the impact of the changes. So, still considering PUT as method, and because the API is based on the exchange between client and server of the whole Traffic Influence resource, are you suggesting to improve the documentation to specify which parameter can be modified in the PUT and which one are not? Or are you suggesting a revision of the code with the PUT not using the TI Resource?

FabrizioMoggio:
@gunjald You suggestion was actually great. I have changed the PUT method into PATCH and I have defined two new resources, derived from the original TrafficInfluence Resource, for the method POST and PATCH. This two new resources just have the editable parameters allowed for those methods.

gunjald:
> @gunjald You suggestion was actually great. I have changed the PUT method into PATCH and I have defined two new resources, derived from the original TrafficInfluence Resource, for the method POST and PATCH. This two new resources just have the editable parameters allowed for those methods. Please, have a look on the new version.

I think the PATCH methods added really helps in this form. I have two observations though,
1. The reference to "Partner OP" in the statement "Unique identifier generated by the partner OP to identify an instance of the application on a specific zone." Do you only mean to say "OP" and not the "Partner OP"?
2. In Patch request the one of the modifiable parameter is "Device" which could be MSISDN or the IP Address. For the modification purpose usage of the Device can result into different conditions e.g. that the POST request contains the MSISDN while the PATCH contains the IP Address or vice versa. So do we really want to support such situations is there is something missing in my understanding?

FabrizioMoggio:
@gunjald 

1. I just copied the InstanceIdentifier definition from the EdgeCloudApi: https://github.com/camaraproject/EdgeCloud/blob/main/code/API_definitions/Resource%20Management/EdgeCloudApi_v0.0.5.yaml.    I don't have any problem changing the text, it is important to have the the APIs aligned in the attributes definition though.  
2. I don't see any problem changing the Device type, it can be totally different, maybe even another user. Which kind of problem  do you foresee?

sergiofranciscoortiz (dismissed):
Approved after explanations in CAMARA group call, although there is not a use case identified at the moment that requires to change the user details, but it makes the definition more simple and this is not expected to cause any issues, anyway thiscould be reviewed in future releases.

FabrizioMoggio:
@gunjald I have fixed the typo related to your request for changes. Can you verify and approve it?

FabrizioMoggio:
@gunjald about the parameter "applicationId" in the TI API. 

That parameter is provided by the EdgeCloudAPI. 

I just noticed that the definition for that parameter is changed in the current version of the EdgeCloudAPI. 

Now it is "AppIdentifier". Am I right? 

If so we should align the TI API code.

gunjald:
> @gunjald about the parameter "applicationId" in the TI API.
> 
> That parameter is provided by the EdgeCloudAPI (https://github.com/camaraproject/EdgeCloud/blob/main/code/API_definitions/Resource%20Management/EdgeCloudApi_v0.0.5.yaml).
> 
> I just noticed that the definition for that parameter is changed in the current version of the EdgeCloudAPI.
> 
> Now it is "AppIdentifier". Am I right?
> 
> If so we should align the TI API code.
I think lets keep it as is for now. We are still having more alignments based on CAMARA API design guidelines and some of the definitions may undergo change. Till that point i would suggest to keep it as is and once all contribution for resource management and orchestration will be updated then we can take this up at that point.

FabrizioMoggio:
about the parameter "applicationId", I totally agree. 

Anyway this version (v0.9.1) is not relevant anymore and we can close the PR

PR #111, opened 2023-07-06
Update Simple Edge Discovery flows

Kevsy:
Clarified the first scenario and added and additional scenario for Simple Edge Discovery

### DIFF
--- a/documentation/SupportingDocuments/Harmonisation of APIs/describing and harmonising the Edge APIs.md
+++ b/documentation/SupportingDocuments/Harmonisation of APIs/describing and harmonising the Edge APIs.md
@@ -94,30 +94,32 @@ Abbreviations used for API names:
 
 Constraints:
 - Network: cellular (4G/5G)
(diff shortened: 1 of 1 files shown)


crissancas (changes requested):
Kevin, I would add an specific comment to emphazise that the interface used is the NBI.

Additionally, I think there is no need for a strict dependency on the hyperscaler account, I mean, our idea is also to use this API with GSMA Telco Edge solution. Could you explain why is needed a hyperscaler account?

Kevsy:
@crissancas thanks, changes made and committed to this PR - please can you review?

crissancas:
Thanks Kevin!

Kevsy:
Re-opened as it was closed without merge - @crissancas please can you merge to main? Thanks!

crissancas (approved):

crissancas:
Sorry Kev! My mistake, I didn't realize.

PR #120, opened 2023-07-25
Create EDS_API_Test.feature

Kevsy:
This PR adds a Gherkin `feature` file with tests for Simple Edge Discovery, as required by the Commonalities [API Readiness Checklist](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-Readiness-Checklist.md)

### DIFF
--- a/code/cucumber/src/test/resources/feature/EDS_API_Test.feature
+++ b/code/cucumber/src/test/resources/feature/EDS_API_Test.feature
@@ -1,0 +1,48 @@
+#/*- ---license-start
+#* CAMARA Project
+#* ---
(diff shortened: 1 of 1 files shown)


gunjald (changes requested):
For the EDSInvalidUEIdentity, can the error code "404 Not Found" be used instead of internal server error ?

Kevsy:
@gunjald on second thoughts, to align with Device Identifers and Quality on Demand APIs: I've made the 'EDSInvalidUEIdentity' a `404` as suggested above. It's technically not ideal for the reasons given, but at least it wil be consistent across CAMARA so probably the best option. I've update the .feature file accordingly (same Pull Request)

Kevsy:
Superseded by #201

PR #122, opened 2023-08-06
Alignment of the simple edge discovery API as per the CAMARA commonality group guidelines

gunjald:
Contribution from 6G-XR (https://www.6g-xr.eu/) towards the alignment of the simple edge discovery API as per the CAMARA commonality group guidelines for the REST API

### DIFF
--- a/code/API_definitions/simple_edge_discovery.yaml
+++ b/code/API_definitions/simple_edge_discovery.yaml
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.0.3
 info:
   title: Simple MEC Discovery API
   version: 0.8.1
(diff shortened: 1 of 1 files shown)


gunjald:
Accidentally raised the request hence closing.

gunjald:
Contribution from 6G-XR (https://www.6g-xr.eu/) towards the alignment of the simple edge discovery API as per the CAMARA commonality group guidelin](https://github.com/camaraproject/EdgeCloud/pull/122)

Kevsy:
Thanks @gunjald - in parallel I've been doing the same exercise, namely to align the data types and error types with those used in other CAMARA APIs (in particular the various device identifiers used in QOD and Device Status) and to ensure alignment with Commonalities API style guide. I also propose to simplify the API by removing the unused query parameters and datatypes (zone, region) which are rather part of the 'full' discovery API.

So: I propose I take the guidance from the commit here and look to combine it with my new version (the goal is the same).

Kevsy:
This should be fixed with #127 which aligns the API with Commonalities

PR #125, opened 2023-08-08
Draft models for GSMA/EdgeXR and 5GFF

Kevsy:
Changes made poer previous PR review (previous PR was removed following branch rename)


Kevsy:
Hi @ThomasEdgeXR , please can you review and merge if approved? Thanks!

ThomasEdgeXR:
Hi @Kevsy i looked at the slides, i'm not fully sure what is the meaning of "All functional intents" on slide 2 between UNI API and Operator. If there are three groups of intents like on slide 3 (Registration, Lifecycle and Disovery) i think it would be more acccurate to assign "Disovery intent" to the link between UNI API and Operator platform and "Registration" and "Lifecycle" to the link between developer and operator platform. 

Also i'm not fully sure on the dotted line to the hyperscaler, does this depict the developer can use some hyperscaler "on the side" (which is of course correct)? If so i believe it should be then all three intent groups. More importantly, there should be a dotted link from the operator platform to hyperscaler as the operator platform can orchestrate both own telco edge as well as hyperscaler resources. 

Does that make sense? Maybe we better transfer this diagrams into mermaid then it's easier to review? If you see any value in that happy to do that.

Kevsy:
Now moved to Discussion #248

Issue #124, opened 2023-08-08
Change project structure to reflect Edge Cloud API 'family'


Kevsy:
Per recent calls, the Edge Cloud APIs will be divided into four groups:

Discovery
Resource Management
Workload Orchestration
Traffic Influence

Each would work on APIs that deliver a set of intents: so Discovery would provide APIs that meet the Discovery intents, and Traffic Influence would contain the Traffic Influence API.   “Resource Management” and “Workload Orchestration” would include intents that are today called ‘runtime’ and ‘provisioning’ and are included in the EdgeXR/GSMA/5GFF contributions.

This  means we can more quickly publish the first API from the Discovery group (‘Simple Edge Discovery’) with a route to extend it to the additional Discovery functionality (‘MEC Exposure and Experience Management’). Both ‘self-build’ and ‘hyperscaler partner’ operators can utilise these APIs.

All operators can support the discovery APIs.

Operators supporting the ‘self-build’ MEC model (i.e. no hyperscaler) can utilise the ‘Resource Management’  and ‘workload orchestration’ groups.

Operators supporting the ‘hyperscaler partner’ MEC model can utilise the Resource Management’  and ‘workload orchestration’ groups if their hyperscaler partner(s) support an integration to do so.

 

That can help operators start to expose the APIs that support their MEC implementation. Otherwise we will find it very difficult to arrive at one single API that everybody can implement.

Each group would be a sub-project containing members that can work on that particular API. All APIs would need to align with the [Commonalities API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md).

Kevsy:
PR #126

PR #126, opened 2023-08-08
Directory restructure per issue 124

Kevsy:
Created new project structure with subdirectories under API Definitions, per #124


crissancas (approved):
Thanks Kevin for the proposal

PR #139, opened 2023-09-07
Update README.md for TI API

FabrizioMoggio:
Completed the readme file for the TI API


gunjald (commented):
Few suggestions,
1. In following starting statement "to enhance the quality of experience leveraging on local instances of and application" I think you want to say "an application" not the "and application"
2. Are the terms "better latency" or "best latency" actually provide the hint on what the API is offering or lets rephrase in a manner to indicate "minimize end to end latency"
3. The last statement about intent may possibly indicate an objective term instead of the term "quality" as the intent as that remains a subjective term in my opinion. We may again depend on objective terms like minimum latency or equivalents that are measurable

gunjald (approved):
Looks good

Issue #145, opened 2023-09-20
Simple Edge Discovery: IP port is not required


Kevsy:
Category: documentation

Documentation incorrectly states thatt if an IP address is used as an identifier, then the port must be provided: which is not correct, as the IP address is simply used to determine the issuing operator.

Kevsy:
Fixed in #146

PR #146, opened 2023-09-20
Minor patch (OAS version +documentation correction)

### LINKED ISSUE #144: Simple Edge Discovery OAS version should be 3.0.3
Per [Commonalities API design guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#11-definition-in-openapi) the OAS version mst be 3.0.3. Simple Edge Discovery 0.9.1 is OAS 3.0.0

Kevsy:
Fixes #144 and #145

### DIFF
--- a/code/API_definitions/Discovery/simple_edge_discovery.yaml
+++ b/code/API_definitions/Discovery/simple_edge_discovery.yaml
@@ -1,7 +1,7 @@
-openapi: 3.0.0
+openapi: 3.0.3
 info:
   title: Simple Edge Discovery API
-  version: 0.9.1
+  version: 0.9.2
   description: |
     # Find the closest MEC platform
     ---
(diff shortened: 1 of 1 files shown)


Kevsy:
> So if the Port is optional and the MSISDN is not mandatory then does it mean the API is not specific to a subscriber but the result could be also valid for other devices who may be assigned the same public IP but different port?

Correct, because the goal is not to identify the subscriber: it's to identify the closest MEC to the terminal. The public IP address subnet indicates which of the network Packet Gateways issued it, and since the operator knows which packet gateway is nearest to each MEC, that is sufficient to answer the API call.

Kevsy (commented):
Added fix to issue #148

crissancas (approved):

PR #154, opened 2023-11-24
EdgeCloud LifecycleManagement API proposal

TEF-RicardoSerr:
Hello everyone!

We propose this API definition as a first contribution to the EdgeCloud Lifecycle Management. This includes:

-    **Application Management** - Deploy and remove applications within artifacts
-    **Application Artifacts Management** - Management of application descriptors, binaries, charts and packages. The format of the artifact definition is based on the way the current Cloud providers (AWS, Azure, Google, etc.) are requesting for the creation of this element
-    **MEC Platfform information** - Retrieves all the Edge nodes available according to some defined parameters. Is to be discussed the way to incorporate the Edge Selection API. As there is an existing contribtion we did not include that query-response until we accept some points to be discussed in the meeting

The API is based also in the contributions done by other partners on this CAMARA group. We can further discuss this file in the next meetings.

Contributors: @javierlozallu, @crissancas, @sergiofranciscoortiz 

Cheers!


code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+  
+    (FAQs will be added in a later version of the documentation)
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: sp-edc@lists.camaraproject.org
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 1.0.0
```
Kevsy, line 43: Version should be 0.1.0 for initial contribution

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
+openapi: 3.0.3
+info:
+  title: Edge Cloud API
+  description:  |
+
+    Edge Cloud API for artifact managment, application management and resource management. 
+    Edge Cloud API for artifact managment, application management and resource management.
```
Kevsy, line 7: Duplicate line

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+  title: Edge Cloud API
+  description:  |
+
+    Edge Cloud API for artifact managment, application management and resource management. 
+    Edge Cloud API for artifact managment, application management and resource management. 
+    # Introduction
+    APIs defined in this version of the specification can be categorized into the following areas:
+    * __Application Management__ - Deploy and remove applications within artifacts
+    * __Application Artifacts Management__ - Management of application descriptors, binaries, charts and packages 
+    * __MEC Platfform information__ - Retrieves all the Edge nodes available according to some defined parameters
```
Kevsy, line 12: s/Platfform/platform

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+              example:
+                status: 409
+                code: CONFLICT
+                message: "Artifact with a running application cannot be deleted"
+        '500':
+          $ref: '#/components/responses/500'
+        '503':
+          $ref: '#/components/responses/503'
+
+  /edgeCloudPlatform:
```
Kevsy, line 247: This duplicates what is in the MEC Exposure and Experience Management API, is that correct?

Kevsy (changes requested):
See comments for minor changes (and one requested clarification)

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: It tends to comply with this intent

“I can retrieve a list of the operator’s MECs and their status, ordering the results by location and filtering by status (active/inactive/unknown)” 5SED 5MEE EXRA
[Developer Intents - Provisioning Intents 1.](https://github.com/camaraproject/EdgeCloud/blob/53aaba1b2f5acc1ba6ecdc8d7d3873f0cc222699/documentation/SupportingDocuments/Harmonisation%20of%20APIs/describing%20and%20harmonising%20the%20Edge%20APIs.md#L85:~:text=%E2%80%9CI%20can%20retrieve%20a%20list%20of%20the%20operator%E2%80%99s%20MECs%20and%20their%20status%2C%20ordering%20the%20results%20by%20location%20and%20filtering%20by%20status%20(active/inactive/unknown)%E2%80%9D%205SED%205MEE%20EXRA)

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+    description: Manage of Artifacts
+  - name: Mobile Edge Cloud Platform
+    description: Mobile Edge Nodes Available
+
+paths:
+  /app:
+    post: 
+      tags:
+        - Applications
+      summary: Deploy an application
```
nicolacdnll, line 75: Deploy and Provision terms are use interchangeably. For clearity, I would suggest to pick one and stick to it.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+  - name: Mobile Edge Cloud Platform
+    description: Mobile Edge Nodes Available
+
+paths:
+  /app:
+    post: 
+      tags:
+        - Applications
+      summary: Deploy an application 
+      description: Ask the operator to provision an application to one or several Edge Application Servers taking into account resources (e.g. vCPU, Memory, network interfaces, storage, GPU)
```
nicolacdnll, line 76: Always for clarity, we might want to also enforce the distinction between Application and Application instance. I see that the document already does it, but partially. But app for discussion.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+        clientCredentials: 
+          tokenUrl: '/oauth2/token'
+          scopes:
+            nbi-mgmt: Access to the APIs
+  
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier assosiated with the artifact. Originating OP generates this identifier when artifact is submitted over NBI.
```
nicolacdnll, line 301:
```suggestion
      description: A globally unique identifier associated with the artifact. Originating OP generates this identifier when artifact is submitted over NBI.
```

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
+openapi: 0.1.0
```
nicolacdnll, line 1: Should this be 3.0.0 or 3.0.1?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+                    
+        descriptorType:
+          type: string
+          enum:
+            - HELM
+            - TERRAFORM
+            - ANSIBLE
+            - SHELL      
+            - COMPONENTSPEC
+          description: Type of descriptor present in the artifact.  App provider can either define either a HELM chart or a Terrform script or container spec.
```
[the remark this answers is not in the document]
nicolacdnll: This would also mean that component spec goes separately. Which also makes sense, becuase the same VM could be deployed using different flavours based on HW or SLAs. That is, VM `a` could be deployed with `2 CPUs, 2 GB DRAM` but also with `8 CPUs, 8 GB DRAM` if it's required.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+tags:
+  - name: Applications
+    description: Manage of Applications
+  - name: Application Artifacts 
+    description: Manage of Artifacts
+  - name: Mobile Edge Cloud Platform
+    description: Mobile Edge Nodes Available
+
+paths:
+  /app:
```
gunjald, line 71: The /app endpoints seems to be provisioning an application. But the application itself is not defined. Though it appears that the artefact has been treated as application. So i would propose to make some changes to make the parameters name consistent and if they refer to a single functionality then make their name same. Could be either of artefact or application.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+    post:
```
gunjald, line 72: As discussed in the group meeting we should also have a GET method to retrieve the provisioning status of the application.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+                message: "Application already deployed"
+        '500':
+          $ref: '#/components/responses/500'
+        '501':
+          $ref: '#/components/responses/501'
+        '503':
+          $ref: '#/components/responses/503'
+
+  /app/{appId}:
+    delete:
```
gunjald, line 119: As discussed in the meeting there should be a GET method also to retrieve the termination status of the app instance.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,703 @@
...
+          scopes:
+            nbi-mgmt: Access to the APIs
+  
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier assosiated with the artifact. Originating OP generates this identifier when artifact is submitted over NBI.
+       
+    AppProviderId:
```
gunjald, line 303: Do we really need an explicit appProviderId? As the API is to be authorized by oAuth2.0, the oAuth credentials should be able to determine the identity of the app provider.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: That's right! for now it seem to be the same entities and might be a bit confusing. Maybe a further documentation could be provided within the yaml. Anyhow it is planned that in future definitions we would have three separated entities:
-  Application 
- Application instance
- Artifact
Thats why we as Telefonica, believe it might be better to keep it segregated for now

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree with that. Parameter removed.

[Commit](https://github.com/camaraproject/EdgeCloud/pull/154/commits/3ffe30c6c302393e4c9a0850bc2879ad5e8bf6e5)

TEF-RicardoSerr (commented):
Review finished

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+              operatingSystem:
+                $ref: '#/components/schemas/OperatingSystem'
+              cpuArchitecture:
+                $ref: '#/components/schemas/CpuArchType'
+              images:
+                description: List of all images assosiated with the component. Images are uploaded or specified using UploadFile apis
+                type: array
+                items: 
+                  $ref: '#/components/schemas/FileId' 
+                minItems: 1
```
gainsley, line 401: There are no fileIds to associate with the component spec, as the file is actually part of the Artifact now. I suggest you remove the images field here, and change the component spec to a single ref instead of an array. That way it's clear there is a single component spec associated with the file that is specified in the Artifact.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+          $ref: '#/components/schemas/RegionId'
+      required:
+        - artifactId
+        - appName
+    AppName:
+      type: string
+      description: Name given by the application provider
+    ArtifactId:
+      type: string
+      pattern: 6sa85a64-5117-3562-c3fd-2c913a66aba1
```
gainsley, line 322: pattern should be a regular expression, or change it to "format: uuid".

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+      type: string
+      format: uuid
+      description: A globally unique identifier assosiated with the image file. Originating OP generates this identifier when file is uploaded over NBI.
+ 
+    MecDetails:
+      type: object
+      required:
+        - geolocation
+      properties:
+        edgeResourceName:
```
gainsley, line 516: I suggest renaming the "edgeResourceName" field to just "name", as it's already scoped within the MecDetails object. This is consistent with other objects that just have a "name" field.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+            - x86
+          example: x86_64  
+        family:
+          type: string
+          enum:
+            - RHEL
+            - UBUNTU
+            - COREOS
+            - WINDOWS
+            - OTROS
```
gainsley, line 546: I think OTROS should be OTHER?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+            - OS_LICENSE_TYPE_ON_DEMAND
+ 
+    RegionId:
+      type: string
+      description:  Human readable name of the geografical zone of the MEC. Defined by the OP.
+      enum:
+        - North
+        - South
+        - West
+        - East
```
gainsley, line 569: Doesn't seem like this should be an enum, it should be left as a plain string without limiting what values it can be.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+    APIs defined in this version of the specification can be categorized into the following areas:
+    * __Application Management__ - Deploy and remove applications within artifacts
+    * __Application Artifacts Management__ - Management of application descriptors, binaries, charts and packages 
+    * __MEC Platform information__ - Retrieves all the Edge nodes available according to some defined parameters
+    # Relevant terms and definitions
+    This section provides definitions of terminologies commonly referred to throughout the API descriptions.
+ 
+    * __Application Provider__ - An application developer, onboarding his/her edge application on operator platform (MEC).    
+    * __Artifact__ - Descriptor, binary image, charts or any other package assosiated with the application.
+    * __OP__ - Operator Platform.
```
gunjald, line 17: Should we mention OP is taken from the GSMA OPG.02?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+    * __Region__ - Human readable name of the geografical zone of the MEC. Defined by the OP.
+    # API Functionality
+    __Application Management__
+    * __onboardApplication__ - Submits an application details to an OP. Based on the details provided,  OP shall do bookkeeping, resource validation and other pre-deployment operations. 
+    * __removeApplication__ - Removes an application from an OP
+ 
+    __Application Artifacts Management__
+    * __onboardArtifact__  Submits an artifact to operator platform
```
gunjald, line 25: Should we define what is an artifact for clarity and its relation with Application?

code/API_definitions/EdgeCloud_LcM.yaml
gainsley: My preference would be to leave user/roles/permissions management APIs out of this spec, as every operator will have their own approach to this. The intent here is to make it possible for such a management system to be able to manage separate permissions for groups of objects (i.e. a Project that consists of apps, app instances, etc). This is a pretty standard approach.

gunjald: I agree with the roles or permission management part which can be out of this spec. Also, i would suggest to keep the appProviderId for future consideration and keep the API bit simple in first version unless there are very compelling reason to keep it.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+          description: CPU Instruction Set Architecture (ISA) E.g., Intel, Arm etc. 
+    ProvisionedApp:
+      description: Information about the provisioned application
+      type: object
+      properties:
+        appInstanceId:
+          $ref: '#/components/schemas/AppInstanceId'
+        edgeResourceName:
+          $ref: '#/components/schemas/EdgeResourceName'
+    EdgeResourceName:
```
gunjald, line 486: If EdgeResourceName refer to same parameter as used in "Simple Edge Discovery" request then we need to assess how to make a single definition for such common available across multiple APIs.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+tags:
+  - name: Applications
+    description: Manage of Applications
+  - name: Application Artifacts 
+    description: Manage of Artifacts
+  - name: Mobile Edge Cloud Platform
+    description: Mobile Edge Nodes Available
+ 
+paths:
+  /app:
```
gunjald, line 66: Shouldnt we keep some AppName or AppId in the path for app provisioning e.g., /app/{appId} instead of keeping AppName in request body?

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: We find this issue needed to be discussed in the next meeting as the neccesity to differentiate the Artifact from the application may not be trivial. Our point of view is that an app may need of multiple artifacts to work so maybe a future chang that might be needed in our definition is that the application would accept an array of `artifactIds`. Anyhow in GSMA the artifact and application refers to a unified element (_edgeApplicationManifest_) as you are proposing. Also in the hyperscalers this applies, they usually query a single template of the service to be deployed including the app instance and the artifacts in the same json request.

We can discuss it!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Regarding your last comment, to cover that we proposed the [PR #159](https://github.com/camaraproject/EdgeCloud/pull/159). This way it would be retrieved the more optimal edge where an instance of the application is deployed, including the URI to access it.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: It can be considered once the discussion regarding single app+artifact vs separated is resolved. Our view is that if it finally separated it would be better to keep it this way so to have better sinergy within artifact and application methods.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,658 @@
...
+          description: Identificator of the application generated by the OP NBI once the provisioning was successful, terminate all the instances related to this appId
+          required: false
+          schema:
+            $ref: "#/components/schemas/AppId"
+        - name: appInstanceId
+          in: query
+          description: Identificator of the specific application instance that will be terminated
+          required: false
+          schema:
+            $ref: "#/components/schemas/AppInstanceId"
```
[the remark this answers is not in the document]
TEF-RicardoSerr: It was defined this way with this rationale:
-  If a user would delete an appId they will be deleting all the appInstances asociated to it
- If a user wanted to delete only a appInstance, they could make the delete request including app instance in the request body

If finally artifact and app are unified as a same entity this could also simplifyed. So we suggest to wait until that discussion is resolved.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree @gunjald , for this fisrt MVP it is assumed that the authentication has already been done outside the API scope. Anyhow we find this usefull and could be integrated in future versions.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: AppProviderId is no longer being used as discussed above. Anyhow ArtifactId has been modified

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, updated.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: The idea is to incorporate simple edge discovery in this set of request as it is part of the LCM of Edge Cloud. We yet belive that refering to `edgeResourceName` is more aproppiate than the existing one in the simple edge discovery. As the `edgeResource` could be refering not only to the name of the resource but also to the resources existing on a specific node. This is because in the simple edge discovery API it is not being refered the contain of the resources but only the resources itselves.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: We suggest to have it as it is so to clarify which name the response is refering to (there are more schemas and responses using parameters like name). Also as this is used in the response of a GET method, we find it more clarfying. But we can discuss it.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Thats right!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, updated!

TEF-RicardoSerr (commented):
Review of the proposed changes and discussions

code/API_definitions/EdgeCloud_LcM.yaml
gainsley: I agree, for us an Artifact is the actual binary data, i.e. POSTing an artifact is in fact uploading a binary file to the platform, not storing a reference.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+                      description: Name of the network.  In case the application has to be assoisated with more then 1 network then app provider
+                                          must define the name of the network on which this interface has to be exposed.  This parameter is required only if 
+                                          the port has to be exposed on a specific network other then default.
+                    protocol:
+                      type: string
+                      enum:
+                        - TCP
+                        - UDP
+                        - HTTP/HTTPS
+                      description: Defines the IP transport communication protocol i.e., TCP, UDP or HTTP
```
nicolacdnll, line 429: Note that:
- HTTP(s) aren't transport protocols.
- If we want to extend this to app protocols, I would suggest we leave room for less used protocols using an `ANY` option.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+            - OTHERS
+        version:
+          type: string
+          enum:
+            - OS_VERSION_UBUNTU_2204_LTS
+            - OS_VERSION_RHEL_8
+            - OS_VERSION_RHEL_7
+            - OS_VERSION_ DEBIAN_11
+            - OS_VERSION_COREOS_STABLE
+            - OS_MS_WINDOWS_2012_R2
```
nicolacdnll, line 550: Is this really needed? I guess we will end up with a very long list.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+              - operatingSystem
+              - cpuArchitecture
+              - images
+              - networkInterfaces
+              - numOfInstances
+              - restartPolicy
+            properties:
+              componentName: 
+                type: string
+                pattern:  ^[A-Za-z0-9][A-Za-z0-9_]{6,62}[A-Za-z0-9]$
```
nicolacdnll, line 385: I think we need something like this `^[a-zA-Z]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$` for RFC 1035 (stole it from [here](https://stackoverflow.com/questions/56835571/check-domain-names-according-to-rfc-1035-standard-in-java)). 
Hyphens are accepted, underscores are not, and the max is 63 (1+61+1).

nicolacdnll: And, does it have to be a valid RFC 1035 here?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+        '501':
+          $ref: '#/components/responses/501'
+        '503':
+          $ref: '#/components/responses/503'
+ 
+  /app/{appId}:
+    delete:
+      tags:
+        - Applications
+      summary: Terminate an Application
```
gunjald, line 122: Just to be technically correct we can say "Terminate an application instance"

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+      description: Terminate the running instance of an application in one or several Edge Application Servers
+      operationId: deleteApplication
+      parameters:
+        - name: appId
```
gunjald, line 126: I haven't seen using query parameters in DELETE operations. Shouldn't both the appId and  appInstanceId be path parameters? E.g., /app/{appId}/instance/{appInstanceId}

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+          in: query
+          description: Identificator of the application generated by the OP NBI once the provisioning was successful, terminate all the instances related to this appId
+          required: false
+          schema:
+            $ref: "#/components/schemas/AppId"
+        - name: appInstanceId
+          in: query
```
gunjald, line 133: My suggestion based on typical usage pattern seen is to make it a path parameter.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+          description: Identificator of the specific application instance that will be terminated
+          required: false
+          schema:
+            $ref: "#/components/schemas/AppInstanceId"
+      responses:
+        '204':
```
gunjald, line 139: We may have HTTP status code 202 also defined as DELETE may not be synchronous operation

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+      requestBody:
+        description: An application can consist of multiple components. App providers are allowed to define seperate artifacts for each component  
+                            or  they could define a consolidated artifact at application level. 
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ArtifactDetails'
+        required: true
+      responses:
+        '201':
```
gunjald, line 170: we may have HTTP 202 also added as this operation may not be an immediate operation. And a GET method defined to retrieve the actual result later.

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: As simple edge discovery (SED) API may become part of this file itself, we may need to harmonize that if the conceptual meaning of edgeResourceName for this API is same as in SED then we should have a one definition of this parameter

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: I think if in the first version we only support containers type applications then we can possibly handle it better

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+        appName:
+          $ref: '#/components/schemas/AppName'
+        edgeResourceName:
+          $ref: '#/components/schemas/EdgeResourceName'
+        regionId:
+          $ref: '#/components/schemas/RegionId'
+      required:
+        - artifactId
+        - appName
+    AppName:
```
gunjald, line 314: we may refine this in terms of valid values for the AppName string including minLength and maxLength and patterns for valid chars

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+      description: Name given by the application provider
+    ArtifactId:
+      type: string
+      format: uuid
+      description: A globally unique identifier assosiated with the artifact. Originating OP generates this identifier when artifact is submitted over NBI.
+ 
+    ArtifactDetails:
+      properties:
+        name:
+          type: string
```
gunjald, line 325: we may refine this in terms of valid values for the name including minLength and maxLength and patterns for valid chars

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+                code: CONFLICT
+                message: "Application already provisioned"
+        '500':
+          $ref: '#/components/responses/500'
+        '501':
+          $ref: '#/components/responses/501'
+        '503':
+          $ref: '#/components/responses/503'
+ 
+  /app/{appId}:
```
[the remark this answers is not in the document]
gunjald: To me 2nd option is preferred, "change the in: path and set it to required"

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+          - inactive
+          - unknown   
+        default: unknown 
+    UploadedArtifact:
+      description: Information about the uploaded artifact
+      type: object
+      properties:
+        artifactId:
+          $ref: '#/components/schemas/ArtifactId'   
+    Uri:
```
gunjald, line 575: We need to define the Uri format to enable validation on user input

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,648 @@
...
+      type: string
+ 
+    ErrorInfo:
+      type: object
+      properties:
+        status:
+          type: integer
+          description: HTTP status code returned along with this error response
+        code:
+          type: string
```
gunjald, line 492: Need to define the string format for code possibly with a regular expression e.g. sequence of digits

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: I agree we should leave room for application level protocols

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: In the just submitted v0.2.0 of the API, `edge-cloud-nodes` is the name we have selected for this parameter oriented by the GSMA naming. It is open to discussion as the final objective is to have the same names for the Simple Edge Discovery and for this set of methods. It is important to follow CAMARA guidelines (reason why we avoid using telco acronyms like mec).

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: In v0.2.0 in general we try to use the words instantiation (for instances) and submition (for applications)

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: As discussed in the previous CAMARA - Edge Cloud meeting (12th - Dec) we agree to combine Artifact and Application into a single type, included in this new version. 

We clarified the difference between Application and Application Instance, defining two methods. Application is HOW to deploy meanwhile Application Instance is WHERE to deploy it. v0.2.0

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: As discussed in the previous CAMARA - Edge Cloud meeting (12th - Dec) we agree to combine Artifact and Application into a single type, included in this new version. 

We clarified the difference between Application and Application Instance, defining two methods. Application is HOW to deploy meanwhile Application Instance is WHERE to deploy it.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: We finally include this GET methods for app and appinstance in the last v0.2.0 
It has been included a status parameter for app instance including the following possible states:
          - ready
          - instantiating
          - failed
          - terminating
          - unknown

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: GET appInstance method has been included and the response includes URI and status (also the POST method does). We believe that the Discovery section of the API should also include this (item was under discussion during the meeting but was not closed)

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: We have rename the entity of edgeResourceName for edgeCloudNodeName, this is something we have to discuss to harmonize common entities in TI, SED and MVP API.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, updated!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: In the v0.2.0 we defined this entities:

- **Application** - The Application Provider submit application metadata to the Operator Platform.

- **Application Instance** - An Operator Platform instantiate an Application on an Edge Cloud Node when the Application Provider resquest it.

And as discussed, artifact is no longer used

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: AppName is no longer used in v0.2.0

code/API_definitions/EdgeCloud_LcM.yaml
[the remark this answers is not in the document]
TEF-RicardoSerr: In v0.2.0 we updated this method and used both path and required

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, `ANY` parameter was added

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: For now option `OTHER` has been added and some options have been kicked out to reduce the list

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Patterns are no longer used in the v0.2.0 for simplicity, in version 1.0.0 this specific details could be included

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, updated!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: DELETE appInstance has been simplified so only a single instance can be deleted at a time. This still covers the basic intent.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, updated!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Artifact no longer exists in v0.2.0 included the 202 response for the rest of the DELETE methods

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: appName no longer exists and all patterns have been removed for simplicity. Anyhow this specific details (patterns) can be included in the v1.0.0

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: `ArtifactDetails` no longer exists and all patterns have been removed for simplicity. Anyhow this specific details (patterns) can be included in the v1.0.0

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: All patterns have been removed for simplicity. Anyhow this specific details (patterns) can be included in the v1.0.0

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: All patterns have been removed for simplicity. Anyhow this specific details (patterns) can be included in the v1.0.0

TEF-RicardoSerr:
New commit has been submitted regarding the last meeting we have [(Minutes of meeting)](https://wiki.camaraproject.org/display/CAM/2023-12-12%3A+SP-EDC+Minutes). Also previous comments have been answered. Hope API is ready to be merged!

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+      operationId: appInstantiation
+      requestBody:
+        description: Details regarding where the application should be instantiated
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AppInstantiation'
+        required: true      
+      responses:
+        '201':
```
gunjald, line 86: Shouldn't we have HTTP status "202 Accepted" to indicate app instantiation is in progress? It may not be an immediate response as typically it may longer time. And now we have also the GET method by which a developer can retrieve the status of instantiation

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+        - name: appId
+          description: A globally unique identifier associated with the application. OP generates this identifier when the application is submitted over NBI.
+          in: path
+          required: true
+          schema: 
+              $ref: '#/components/schemas/AppId'
+        - name: appInstanceId
+          description:  A globally unique identifier associated with a running instance of an application. OP generates this identifier. 
+          in: query
+          required: false
```
gunjald, line 139: Do we need to make "required" as true? Otherwise we need to define if no instance id has been provided then what would be the behavior?  To terminate all the instances of this app?

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Still I think we should have 202 supported. As app submission still may not be immediate handling.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree, Included

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: The intention is that all the instances can be deleted if the instanceID is no specificated. This behaviour could be explained in the documentation

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+          description: CPU Instruction Set Architecture (ISA) E.g., Intel, Arm etc. 
+    InstantiatedApp:
+      description: Information about the instantiated application
+      type: object
+      properties:
+        appInstanceId:
+          $ref: '#/components/schemas/AppInstanceId'
+        edgeCloudNodeName:
+          $ref: '#/components/schemas/EdgeCloudNodeName'
+        uri:
```
gunjald, line 537: We should add some description on what the API user can expect from the "Uri" and how it can be used

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+          description: Application instantiation accepted
+          content: 
+            application/json:
+              schema:
+                type: object
+                properties:
+                  appId:
+                    $ref: '#/components/schemas/AppId'
+                  aplicationInstace:
+                    type: array
```
gunjald, line 96: Can we describe if more than one "InstantiatedApp" object can be expected in under some conditions as aplicationInstace is defined as an array? And in that case how should information contained in InstantiatedApp objects can be used?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+                code: CONFLICT
+                message: "Application already instantiated in the given Edge or Region"
+        '500':
+          $ref: '#/components/responses/500'
+        '501':
+          $ref: '#/components/responses/501'
+        '503':
+          $ref: '#/components/responses/503'
+  /appInstance/{appId}:
+    get:
```
gunjald, line 123: Should a query parameter also be provided to filter using edge, region etc?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  appId:
+                    $ref: '#/components/schemas/AppId'
+                  aplicationInstace:
+                    type: array
+                    items:
+                      $ref:  '#/components/schemas/InstantiatedApp'
```
gunjald, line 155: Just a suggestion if we can we can use a more descriptive name e.g. "AppInstanceInfo" etc

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+          schema: 
+              $ref: '#/components/schemas/AppId'
+        - name: appInstanceId
+          description:  A globally unique identifier associated with a running instance of an application. OP generates this identifier. 
+          in: query
+          required: false
+          schema: 
+              $ref: '#/components/schemas/AppInstanceId'     
+      responses:
+        '201':
```
gunjald, line 143: I think a GET response of 201 will not be a valid response code. Please check

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+      operationId: deleteAppInstance
+      parameters:
+        - name: appInstanceId
+          in: path
+          description: Identificator of the specific application instance that will be terminated
+          required: true
+          schema:
+            $ref: "#/components/schemas/AppInstanceId"
+      responses:
+        '202':
```
gunjald, line 183: As the 202 means that the deletion can some time then should there be a corresponding GET to retrieve the result of the DELETE app instance?

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+      operationId: getApp
+      parameters:
+        - name: appId
+          description: A globally unique identifier associated with the application. OP generates this identifier when the application is submitted over NBI.
+          in: path
+          required: true
+          schema: 
+              $ref: '#/components/schemas/AppId'
+      responses:
+        '201':
```
gunjald, line 258: I suppose GET will not have status code of 201. Please check.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+          type: string
+          enum:
+            - VM
+            - CONTAINER
+            - HELM 
+        fileName:
+          type: string
+          minLength:  8
+          maxLength: 32
+          description: Name of the file with the extension eg. zip, targz, etc.
```
gunjald, line 409: It need to be described what is expected from the app developer from the fileName? Which file this element is referring to?

gunjald: It also needs to be described if the fileName has any relation with the repository attribute below? That means the expectation is that OP will try to pull the file(s) from the repository?

TEF-RicardoSerr: The idea we were thinking was that the combination of both fields `FileName` and `url` will let the developer to build the correct path to the binary file in their repository.
Agree this needs to be documented.

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Suggestion accepted!

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Suggestion accepted

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -0,0 +1,712 @@
...
+        '403':
+          $ref: '#/components/responses/403'
+        '409':
+          description: Conflict
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorInfo'
+              example:
+                status: 409
```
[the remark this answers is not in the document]
TEF-RicardoSerr: The idea behind is that if a simple node is parsed and that node exists then that general response applies, same with region

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: It is an array because depending whether there is a filter in the request, the app could be instantiated in multiple edge nodes (eg. Region or all the nodes if no filter is parsed)

code/API_definitions/EdgeCloud_LcM.yaml
TEF-RicardoSerr: Agree! Once the pull request is merged, further documentation can be added

TEF-RicardoSerr (commented):
Deepak comments review

sergiofranciscoortiz (commented):
After several iterations/commits with improvements on the API definition, and as it was agreed in  [12th December Meeting](https://wiki.camaraproject.org/display/CAM/2023-12-12%3A+SP-EDC+Minutes) to advance towards an MVP API, it would help to approve this pull request and merge into main, so further changes proposals ( parameter patterns, descriptions and other "minor" changes) and an eventual merge with Simple Edge Discovery API in a more ordered way in separated PRs.

sergiofranciscoortiz:
@Kevsy , @gunjald ,@FabrizioMoggio As agreed during  [19th December EdgeCloud Meeting](https://wiki.camaraproject.org/display/CAM/2023-12-19%3A+SP-EDC+Minutes), this PR  has evolve to a point where it should be merged as pending conversations could be treated in separated PRs to facilitate management. Please review if there is anything else to consider before approval.

gainsley (approved):
I agree this is good starting point and we can file issues to address outstanding concerns, and address each separately in follow-up PRs.

Kevsy (approved):

PR #159, opened 2023-11-30
Simple edge discovery: Optional parameters addition

TEF-RicardoSerr:
Optional parameters addition.

In the previous contribution, the final user would not know the actual connection details . This proposal enhances the user to connect the application if they know the appId of the instance they want to use. This way: If provided an `appId` an `Uri` of the service endpoint/application instance will be returned within the edge details

### DIFF
--- a/code/API_definitions/Discovery/simple_edge_discovery.yaml
+++ b/code/API_definitions/Discovery/simple_edge_discovery.yaml
@@ -130,7 +130,12 @@ paths:
             enum: [
               "closest"
               ]
-            
+        - name: appId
+          in: query
+          required: false
+          description: Application ID of the user request
+          schema:
+            $ref: '#/components/schemas/AppId'
         - name: IP-Address
           in: header
           required: false
@@ -170,7 +175,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/MecPlatforms'
+                type: object
+                properties:  
+                  mecPlatform:
+                      $ref: '#/components/schemas/MecPlatform'
+                  uri:
+                      $ref: '#/components/schemas/Uri'
+                        
         "400":
           $ref: '#/components/responses/400BadRequest'
         "401":
@@ -208,6 +219,11 @@ components:
           tokenUrl: "{tokenUrl}"
           scopes: {}
   schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier associated with the artifact. Originating OP generates this identifier when artifact is submitted over NBI. If this parameter is provided, Uri of the application instance running in the selected edge is returned.
+      
     MecPlatforms:
       type: array
       items:
@@ -247,6 +263,10 @@ components:
         message: 
           type: string
           description: This parameter appears when there was an error. Human readable explanation specific to this occurrence of the problem
+    Uri:
+      type: string
+      description: URI of Service Endpoint if available where the application instance is deployed. Uri is only returned if the appId is provided
+      format: uri
       
 
 ###################################
--- a/code/API_definitions/Discovery/simple_edge_discovery.yaml
+++ b/code/API_definitions/Discovery/simple_edge_discovery.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Simple Edge Discovery API
-  version: 0.9.3
+  version: 0.9.4
   description: |
     # Find the closest MEC platform
     ---


Kevsy:
HI @TEF-RicardoSerr 

The Simple API can *partly* do this now: the developer learns the closest edge platform, and they also know the URI of their apps deployed on that particular edge. Or, they have not deployed on that closest edge platform, and so they may decide to deploy (or accept connection to a further edge their app is deployed on).

Extending that functionality is complex - because it requires the network being aware of which apps are deployed on which edge platform, and that requires out-of-band provision of that information. So it makes the Simple API no longer Simple :) 

The full functionality is supported today in the MEC Exposure and Experience Management API, and I recommend to work on that API rather than extend the Simple API.

gunjald:
> HI @TEF-RicardoSerr
> 
> The Simple API can _partly_ do this now: the API consumer learns the closest edge platform, and they also know the URI of their apps deployed on that particular edge. Or, they have not deployed on that closest edge platform, and so they may decide to deploy (or accept connection to a further edge their app is deployed on).
> 
> Extending that functionality is complex - because it requires the network being aware of which apps are deployed on which edge platform, and that requires out-of-band provision of that information. So it makes the Simple API no longer Simple :)
> 
> The full functionality is supported today in the MEC Exposure and Experience Management API, and I recommend to work on that API rather than extend the Simple API.

I also have the same opinion as of Kesvy that to not further extend it for now. Though I also agree with the need for newer APIs that can make the SED more useful. For example the API customer can get the information of the closest MEC platforms for an end user but to access and manage any applications on that platform is outside of the SED API. I as an developer may want to further use the MEC platform information to deploy applications via API for a more smoother experience provided I have the commercial agreement with the telco to use the given MEC platforms. So in my opinion we should enlist such intents to derive the need for further extensions or the APIs to provide complete automation of such use cases.

TEF-RicardoSerr (commented):
Comments on the proposed changes

TEF-RicardoSerr:
Our proposals are tended to have a single API definition covering the different proposed intents as expresed in the #154. In this PR proposal, the incorporation of the appId parameter is **optional** so the functionalities proposed in this simple edge discovery definition that you have described are still possible.  This makes our proposal and yours compatible.

Kevsy:
Thanks @TEF-RicardoSerr : My thoughts on adding an optional parameter for appId...

1. I agree this is a non-breaking change/'Backwards Compatible' according to [API Design Guidelines](#52-backwards-and-forward-compatibility) , which is good...

2. ...but I'm not sure how the operator knows which Edge Platform(s) the appId is deployed on. That part is enabled by a series of provisioning operations in the MEC Exposure and Experience Management API, meaning that the operator has a handle on the particular app being queried - but those operations are (by design) not part of the Simple API. So is the idea that the operator has to find their own way to determine which Edge Platform is hosting the appId?

3. What is the proposed response body and status code for operators that do not support appId in the query?

ThomasEdgeXR:
As discussed in the call today i believe discovering the "nearest edge platform" (i.e. the purpose of Simple Edge Discovery) and discovering the "best existing app endpoint" are two entirely different things which should be separate APIs also considering there are called by different users (simple edge discovery by the application provider and app discovery by the UE). Check https://www.verizon.com/business/5g-edge-portal/documentation/verizon-5g-edge-discovery-service.html for clarification of the purpose and workflow of simple edge discovery.

TEF-RicardoSerr:
> Thanks @TEF-RicardoSerr : My thoughts on adding an optional parameter for appId...
> 
> 1. I agree this is a non-breaking change/'Backwards Compatible' according to [API Design Guidelines](#52-backwards-and-forward-compatibility) , which is good...
> 2. ...but I'm not sure how the operator knows which Edge Platform(s) the appId is deployed on. That part is enabled by a series of provisioning operations in the MEC Exposure and Experience Management API, meaning that the operator has a handle on the particular app being queried - but those operations are (by design) not part of the Simple API. So is the idea that the operator has to find their own way to determine which Edge Platform is hosting the appId?
> 3. What is the proposed response body and status code for operators that do not support appId in the query?

Hello @Kevsy.
As we defined in the [LCM proposal](https://github.com/camaraproject/EdgeCloud/blob/main/code/API_definitions/EdgeCloud_LcM.yaml), the appId is generated within the OP context once the app is submited by the AP to the OP. As this appId is later used by the OP to run/deploy the appInstance in the desired sites, the OP knows where the app is being deployed and has this information linked to the appId param (as it was also used in the instantiation process).

On behalf the bullet 3 our idea is the API to respond with the current response body (only the `edgeResourceName`). Anyhow we know is under discussion the different discoveries models in different APIs and so maybe the URL could be returned instead or alternatively.

Kevsy:
Propose to close this as the relationship between Edge_LCM and Simple Edge Discovery has been progressed in #215

TEF-RicardoSerr:
As recomended im closing the PR

PR #188, opened 2024-02-02
Update EdgeCloud_LcM.yaml

gunjald:
The proposed PR aims to simplify and establish relationship between some of the parameters like Uri and component and componentSpec. This changes the AppManifest and provide a description of component and how it relates to componentSpec and what is expected from the API user. The API user know the information of Helm chart or docker-compose internal and what all of them needs to be exposed to external access.

### DIFF
--- a/code/API_definitions/EdgeCloud_LcM.yaml
+++ b/code/API_definitions/EdgeCloud_LcM.yaml
@@ -446,7 +446,9 @@ components:
                 - HTTP_BEARER
                 - NONE
               description:  The credentials can also be formatted as a Basic auth or Bearer auth in HTTP "Authorization" header. 
-              
+            checksum:
+              type: string
+              description: MD5 checksum for VM and file-based images, sha256 digest for containers  
         componentSpec:
           description: Information defined in "appRepo" point to the application descriptor e.g. Helm chart, docker-compose yaml file etc. The descriptor may contain one or more containers and their associated meta-data. A component refers to additional details about these containers to expose the instances of the containers to external client applications. App provider can define one or more components (via the associated network port) in componentSpec corresponding to the containers in helm charts or docker-compose yaml file as part of app descriptors.
           type: array
@@ -487,7 +489,6 @@ components:
                       minimum: 1
                       maximum: 65535
                       description: Port number exposed by the component. OP may generate a dynamic port towards the component instance which forwards external traffic to the component port.
-.
                     visibilityType:
                       description: Defines whether the interface is exposed to outer world or not i.e., external, or internal.If this is set to "external", then it is  exposed to external applications otherwise it is exposed internally to edge application components within edge cloud. When exposed to external world, an external dynamic port is assigned for UC traffic and mapped to the extPort
                       type: string
--- a/code/API_definitions/EdgeCloud_LcM.yaml
+++ b/code/API_definitions/EdgeCloud_LcM.yaml
@@ -421,7 +421,7 @@ components:
           type: object
           required:
             - type
-            - url
+            - imagePath
           properties:
             type:                 
               type: string
@@ -429,7 +429,7 @@ components:
                 - PRIVATEREPO
                 - PUBLICREPO
               description: Application repository and image URI information. PUBLICREPO is used of public urls like github, helm repo etc. PRIVATEREPO  is used for private repo managed by the application developer. Private repo can be accessed by using the app developer provided userName and password. Password is recommended to be the personal access token created by developer e.g. in Github repo. 
-            url:
+            imagePath:
               $ref: '#/components/schemas/Uri'
             userName:
               type: string
--- a/code/API_definitions/EdgeCloud_LcM.yaml
+++ b/code/API_definitions/EdgeCloud_LcM.yaml
@@ -458,7 +458,7 @@ components:
                   required:
                     - interfaceId
                     - protocol
-                    - extPort
+                    - port
                     - visibilityType
                   properties:
                     interfaceId:
@@ -472,12 +472,13 @@ components:
                         - UDP
                         - ANY
                       description: Defines the IP transport communication protocol i.e., TCP, UDP or ANY
-                    extPort:
+                    port:
                       type: integer
                       format: int32
                       minimum: 1
                       maximum: 65535
-                      description: Port number exposed by the component. OP may generate a dynamic port towards the UCs corresponding to this internal port  and forward the client traffic from dynamic port to containerPort.
+                      description: Port number exposed by the component. OP may generate a dynamic port towards the component instance which forwards external traffic to the component port.
+.
                     visibilityType:
                       description: Defines whether the interface is exposed to outer world or not i.e., external, or internal.If this is set to "external", then it is  exposed to external applications otherwise it is exposed internally to edge application components within edge cloud. When exposed to external world, an external dynamic port is assigned for UC traffic and mapped to the extPort
                       type: string
--- a/code/API_definitions/EdgeCloud_LcM.yaml
+++ b/code/API_definitions/EdgeCloud_LcM.yaml
@@ -404,6 +404,7 @@ components:
       properties:
         name:
           type: string
+          pattern: ^[A-Za-z][A-Za-z0-9_]{7,63}$
           description: Name of the application. 
         version:
           type: integer
@@ -414,12 +415,9 @@ components:
             - VM
             - CONTAINER
             - HELM 
-        fileName:
-          type: string
-          minLength:  8
-          maxLength: 32
-          description: Name of the file with the extension eg. zip, targz, etc.
-        repository:       
+        operatingSystem:
+          $ref: '#/components/schemas/OperatingSystem'
+        appRepo:       
           type: object
           required:
             - type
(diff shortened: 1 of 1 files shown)


code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -414,12 +415,9 @@ components:
...
             - HELM 
-        fileName:
-          type: string
-          minLength:  8
-          maxLength: 32
-          description: Name of the file with the extension eg. zip, targz, etc.
-        repository:       
+        operatingSystem:
+          $ref: '#/components/schemas/OperatingSystem'
+        appRepo:
```
gainsley, line 420: With the name "appRepo", this makes it sound like the URL below is the URL for the repository, i.e. docker.io, rather than the URL for the image, i.e. docker.io/hashicorp/httpecho:2.3.1. I suggest to rename the "url" field below to "imagePath", so it is clear the URL is for the image.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -430,114 +428,70 @@ components:
...
+              description:  Username to acces the Helm chart, docker-compose file or VM image repository
             password:
-              type: string      
-              description:  Password to acces the artifact repository
-            token:
-              type: string          
-              description:  Authorization token acces the artifact repository   
+              type: string
+              maxLength: 128
+              description:  Password or personal access token created by developerto acces the app repository
```
gainsley, line 440: It is important to have a checksum field so that when our platform pulls the image, we can verify that our copy is correct and we have pulled the exact image the user intended.

I suggest adding the field:
```
                checksum:
                  type: string
                  description: MD5 checksum for VM and file-based images, sha256 digest for containers
```

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -430,114 +428,70 @@ components:
...
                       type: integer
                       format: int32
                       minimum: 1
                       maximum: 65535
-                      description: Port number exposed by the component. OP may generate a dynamic port towards the UCs 
-                                          corresponding to this internal port and forward the client traffic from dynamic port to containerPort.
-                    interfaceName:
-                      type: string
-                      description:  Interface Name. Required only if application has to be attatched to a network other then default.
+                      description: Port number exposed by the component. OP may generate a dynamic port towards the UCs corresponding to this internal port  and forward the client traffic from dynamic port to containerPort.
```
gainsley, line 480: Naming and comment here is confusing, the field is called "extPort" (external port?) but the comment refers to it as both "internal port" and "containerPort", the latter of which doesn't make sense for VM based deployments. Also "UC" is not defined anywhere, so I'm not sure what that stands for.

I would suggest to rename the field back to "port", and have the comment be:
```
Port number exposed by the component. OP may generate a dynamic port towards the component instance which forwards external traffic to the component port.
```

gainsley (commented):
Thanks for making these changes!

code/API_definitions/EdgeCloud_LcM.yaml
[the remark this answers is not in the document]
gunjald: The token as separate field is removed and the password field is described to be placeholder for the token as well. For example Github or Dockerhub and few other repositories accept password and token for a given user name. So in that case we may have separate fields with either or relationship as I could think. Also, in that case we will need to provide another field for repo type e.g. HTTPS, ssh Git credentials.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -404,6 +404,7 @@ components:
       properties:
         name:
           type: string
+          pattern: ^[A-Za-z][A-Za-z0-9_]{7,63}$
```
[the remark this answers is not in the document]
gunjald: Though I agree that minimum length of 7 chars is subjective but i think there should be some minimum. we can not allow 0 length names just as an hypothetical example. But we can discuss if there is a better proposal for this case.

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: I think this is valid. Will provide a change for this.

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Agree. Can incorporate this change.

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Agree with the observation. Will provide a change for it.

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: But there is one issue now. For VM the imagePath can point to the actual image file. But for the Helm charts or docker-compose file the name imagePath doesn't convey the intent. Also, as there could be multiple containers in the file there needs to be a list of container image name and associated checksums. Any thoughts?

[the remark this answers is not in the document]
gainsley: In most cases where the server accepts either a password or an API key (like Github) with a username, the process is the same in both cases for the client. The OP here is the client, so whether the password field contains a password or an API key, the client doesn't care. It only needs to care if it needs to follow a protocol (like docker login to get a token), or send it formatted as a basic HTTP auth header, or bearer HTTP auth header. I'm fine either way, but if you want to overload the password field I suggest you rename it to "credentials", and then add another field "authType", which could be "docker", "HTTP Basic", or "HTTP Bearer".

code/API_definitions/EdgeCloud_LcM.yaml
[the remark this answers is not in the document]
gainsley: I agree with min length 1 to prevent an empty string. Anything else seems rather arbitrary, though.

code/API_definitions/EdgeCloud_LcM.yaml
gainsley: I agree Helm is tricky because a chart is really a directory with lots of files in it. There's no way to checksum it unless you package it into an archive. But for anything that's a single file (including docker-compose files), the intent is to checksum the file the OP will download. Anything that another program does, like docker parsing the docker-compose and downloading images, or helm parsing the chart files and downloading images, is really under the control of that program and out of our scope (we don't want to overlap with the functionality of docker or helm). So the intent is just to checksum whatever file is pointed to by imagePath. I would suggest that checksum be optional or not applicable for helm charts since it may be a directory structure instead of a single file.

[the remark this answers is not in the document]
gunjald: I think the following looks a good approach "rename it to "credentials", and then add another field "authType", which could be "docker", "HTTP Basic", or "HTTP Bearer"."

gunjald:
I also propose to add a parameter to appRepo to handle Git repositories. Parameter could take input as repoType that could be one value as Git along with other attributes as revision and path to identify the helm chart chart for example.
Any suggestions please..

gainsley:
> I also propose to add a parameter to appRepo to handle Git repositories. Parameter could take input as repoType that could be one value as Git along with other attributes as revision and path to identify the helm chart chart for example. Any suggestions please..

I would suggest to take that up as a separate task from this PR, as its pretty complex and would need some discussion on exactly how it would be handled. For example, what would be the expected file names/directory structure/file types? Would the intent to be to pull a specific branch and commit/tag, or would it be more like a Rancher/ArgoCD CI/CD flow where anytime the branch is updated the OP deploys the new version of the app? My guess would be for each deployment type (docker/kubernetes/helm/vm etc) we'd need to detail to the user exactly what directory and file structure the OP expects to read and how the OP would then attempt to deploy it.

code/API_definitions/EdgeCloud_LcM.yaml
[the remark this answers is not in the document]
gunjald: Changes done and committed to PR

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Changes done and committed to PR

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Changes done and committed to PR

code/API_definitions/EdgeCloud_LcM.yaml
gunjald: Changes done and committed to PR

gunjald:
> > I also propose to add a parameter to appRepo to handle Git repositories. Parameter could take input as repoType that could be one value as Git along with other attributes as revision and path to identify the helm chart chart for example. Any suggestions please..
> 
> I would suggest to take that up as a separate task from this PR, as its pretty complex and would need some discussion on exactly how it would be handled. For example, what would be the expected file names/directory structure/file types? Would the intent to be to pull a specific branch and commit/tag, or would it be more like a Rancher/ArgoCD CI/CD flow where anytime the branch is updated the OP deploys the new version of the app? My guess would be for each deployment type (docker/kubernetes/helm/vm etc) we'd need to detail to the user exactly what directory and file structure the OP expects to read and how the OP would then attempt to deploy it.

Seems fine and we can take this proposal in future.

code/API_definitions/EdgeCloud_LcM.yaml
```diff
@@ -501,43 +488,21 @@ components:
...
         - virtType
-        - fileName
         - repository
         - componentSpec
-    CpuArchType:
-          type: string
-          enum:
-            - x86_64
-            - arm_64
-          description: CPU Instruction Set Architecture (ISA) E.g., Intel, Arm etc.
```
nicolacdnll, line 540: Not sure if we should remove this. We sometimes see container images build for specific arch. I would like to see what others experience is here.

nicolacdnll: Clear, Thanks!

gainsley (approved):
Thanks for iterating through those changes!

gunjald:
As no further comments i propose to merge this PR. Any comments can be addressed over this version as they come.

nicolacdnll (approved):

sergiofranciscoortiz (approved):
Resolved conflicts with main before approving changes as no further comments from others.

PR #205, opened 2024-02-12
Update describing and harmonising the Edge APIs.md

Kevsy:
Corrected text to confirm that EDS can be called from an application server, or by an application client on a device.


gunjald (approved):
Description is fine from my side.

PR #211, opened 2024-02-20
Gherkin Test feature file for EdgeCloud_LCM

sergiofranciscoortiz:
First version of Gherkin test feature file for EdgeCloud_LCM including at lis one test with response ok and one test with response nok for all existing POST,GET,DELETE methods in version 0.9.2 (https://github.com/camaraproject/EdgeCloud/pull/208)

### DIFF
--- a/code/Test_definitions/Edgecloud_LCM_Test.feature
+++ b/code/Test_definitions/Edgecloud_LCM_Test.feature
@@ -1,0 +1,139 @@
+#/*- ---license-start
+#* CAMARA Project
+#* ---
+#* Copyright (C) 2022 - 2023 Contributors | Deutsche Telekom AG to CAMARA a Series of LF Projects, LLC
+#* The contributor of this file confirms his sign-off for the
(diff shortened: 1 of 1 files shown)


gunjald:
For  "@EdgeCloud_LCM_app_instantiation_conflict", it looks like for an application pointed by appId can have at max one instance in a given edge cloud zone of a given region. 

If this is the case it should be mentioned somewhere in the API description to provide clarity to the app developers on the behavior of the API.

gunjald:
"@EdgeCloud_LCM_app_innstance_delete_valid_parameter"

The expectation is that the platform accepts the request after validation of all request parameters but actual termination may take time and may not be immediate.

So in my opinion we may say that the outcome is "And the application instance termination initiated" but not deleted.

gunjald:
"@EdgeCloud_LCM_app_innstance_delete_invalid_parameter" seems to be a valid test but doesnt seems to have any invalid parameter as given in the test name. Please check

PR #224, opened 2024-03-21
Update Traffic_Influence_API_Readiness_Checklist.md

FabrizioMoggio:
the API implementation is not Mandatory

#### What type of PR is this?

* correction
* documentation

to be aligned with: https://github.com/camaraproject/Commonalities/blob/main/documentation/API-Readiness-Checklist.md


FabrizioMoggio:
I don't get why this PR is blocked. It seams to me that MegaLinter detects an error on other yaml files (e.g. MEC exposure and experience management.yaml).

@javierlozallu have you any suggestion? Is there any action I need to do?

javierlozallu (approved):

javierlozallu:
Yes, MegaLinter is controlling the yaml format and we need to adapt the line length to 80 characters. SED and LCM are already adapted, you can check it out. But we have some old YAMLs that need to be adapted if we have to keep them or delete them. But you can merge this PR now that I have approved.

gunjald:
We should have a way to tell mega linter what files are to be checked instead of whole repository. Or we move the files not required in some achieves.

PR #240, opened 2024-05-15
TI API Rel 0.9.4: CR in LF format

FabrizioMoggio:
Carriage Return in LF format.

#### What type of PR is this?

* bug

CR should be in LF format while in Rel 0.9.3 it was in Windows format. I decided not to include this update in rel 0.9.3 otherwise every code line is effected and the modification were not visible anymore.

### DIFF
--- a/code/API_definitions/Traffic_Influence.yaml
+++ b/code/API_definitions/Traffic_Influence.yaml
@@ -5,7 +5,7 @@ openapi: 3.0.3
 ############################################################################
 info:
   title: OPAG-CAMARA Traffic Influence API
-  version: 0.9.3
+  version: 0.9.4
   description: |
     ## Overview
     The reference scenario foresees a Service, composed by one or more Service


javierlozallu (commented):
@FabrizioMoggio I am not familiar with LF format versus Windows format and I don't find any differences with version 0.9.3, could you explain briefly? 
Thank you.

FabrizioMoggio:
You can not see any difference because there are no differences in this 0.9.4 :-).  Rel 0.9.3 was already actually using Carriage Return LF format. I just discovered that if you copy a file from GitHub using the "Copy RAW file" button, the Carriage Return is implemented in Windows format even if the original file has the Carriage Return in LF format. This confused me because I couldn't understand why the file was in Windows format. So I made the patch. So I can close this PR.

Screenshot 2024-05-16 115457

FabrizioMoggio:
About the Image, on the left you can see Carriage Return as LF (Linux) and on the right they are implemented as CR.LF (Windows)

PR #245, opened 2024-05-20
Application Endpoint Discovery First Contribution

crissancas:
#### What type of PR is this?

* enhancement/feature

    The Application Discovery API extends beyond the capabilities of the
    Simple Edge Discovery API by not only locating the nearest Edge Cloud Zone
    but also directly linking to the application endpoints within those zones.
    This PR covers intent 21.

Fixes #212 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/Application_Endpoint_Discovery_API.yaml
+++ b/code/API_definitions/Application_Endpoint_Discovery_API.yaml
@@ -1,0 +1,596 @@
+---
+openapi: 3.0.3
+info:
+  title: Application Endpoint Discovery API
+  version: 0.1.0-wip
+  description: |
+    # Connect to the Nearest Application Endpoint
+    ---
+    # Overview
+    The Application Discovery API extends beyond the capabilities of the
+    Simple Edge Discovery API by not only locating the nearest Edge Cloud Zone
+    but also directly linking to the application endpoints within those zones.
+    This crucial enhancement enables immediate connection to the most relevant
+    service endpoint, optimizing the application's interaction with
+    edge computing resources.
+    # Introduction
+    Operators often host multiple Edge Cloud Zones across various territories.
+    Connecting an application directly to the nearest operational service
+    endpoint minimizes latency because data travels the shortest possible route.
+    Unlike the Simple Edge Discovery API, which only provides the zone location,
+    this API calculates and provides the actual endpoints within the nearest
+    zone. This approach ensures that developers can instantly integrate these
+    endpoints into their applications, improving performance and user
+    experience.
+    After identifying the nearest application endpoint through the API,
+    developers can:
+    * Connect their application client directly to the chosen application
+    endpoint, enhancing the responsiveness and reliability of the service.
+    * Deploy new instances if no suitable endpoints are within the identified
+    Edge Cloud Zone, optimizing their service distribution based on real-time
+    network geography.
+    # Quick Start
+    This API can be utilized by both client applications on user devices and
+    by servers. It uses a straightforward endpoint:
+    `/apps-endpoints?filter=closest`.
+    API consumers are required to authenticate via OAuth2 to obtain access
+    tokens,
+    which must be included in the `Authorization` header of API requests.
+    For accurate service localization, the API requires device-specific
+    identifiers such as:
+    * `IP-Address`: Public IP address of the user device, which can be fetched
+    via
+    external IP discovery services.
+    * `Phone-Number` : The device's phone number in international E.164 format.
+    * `Network-Access-Identifier`: A unique identifier provided by the network
+    operator.
+    If the API is accessed directly from a device within the operator's network,
+    initial requests may be made without explicit device identifiers. However,
+    if this results in a `404 Not Found` error, subsequent requests should
+    include
+    one of the identifiers to ensure accurate localization of the nearest
+    endpoint.
+    ## Example request
+    The following examples demonstrate how to make requests to the Application
+    Discovery API, which provides direct access to the nearest application
+    endpoint
+    within the nearest Edge Cloud Zone based on the user's device identifiers.
+    Example 1: Request with Network-Access-Identifier
+    ```
+    GET /apps-endpoints?filter=closest HTTP/1.1
+    Host: example.com
+    Authorization: Bearer {your_access_token}
+    Accept: application/json
+    Network-Access-Identifier: 4d596ac1-7822-4927-a3c5-d72e1f922c94@domain.com
+    ```
+    Example 2: Request with IP-Address
+    ```
+      GET /apps-endpoints?filter=closest HTTP/1.1
+      Host: example.com
+      Authorization: Bearer {your_access_token}
+      Accept: application/json
+      IP-Address: 84.125.93.10
+    ```
+    Example 3: Request with Phone-Number
+    ```
+      GET /apps-endpoints?filter=closest HTTP/1.1
+      Host: example.com
+      Authorization: Bearer {your_access_token}
+      Accept: application/json
+      Phone-Number: +441234567890
+    ```
+    # Authorization and authentication
+    In order to maintain the integrity and security of the Application Discovery
+    API, access is governed by robust authentication and authorization
+    processes.
+    This ensures that only valid, authorized requests are able to interact with
+    the API and retrieve application endpoint information.
+    ## OpenID Connect
+    The API utilizes OpenID Connect on top of OAuth 2.0 for user authentication.
+    Users must register to obtain client credentials, which are necessary to
+    acquire access tokens from the OpenID provider.
+    ## Obtaining Access Tokens
+    Access tokens are obtained by submitting a request to the designated OpenID
+    Connect URL, utilizing the client credentials. It is imperative that the
+    retrieved token is included in the `Authorization` header as a `Bearer`
+    token for all subsequent API requests.
+    ## Scopes and Permissions
+    During the token request, developers will need to specify the scopes that
+    correspond to the permissions the application requires. Our API strictly
+    enforces scope validation to ensure clients only perform operations that
+    are within their granted permissions.
+    ## Secure Transmission
+    All communication with the API, including token requests, must occur over
+    HTTPS to ensure data confidentiality and protection.
+    ## Error Handling
+    If a request is made without a token or with an expired or invalid token,
+    the API will return a  `401 Unauthorized` status code. Similarly, if the
+    token does not include the proper scopes for the requested operation,
+    a `403 Forbidden` status code will be returned, indicating
+    insufficient permissions.
+    ---
+  contact:
+    email: sp-edc@lists.camaraproject.org
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+externalDocs:
+  description: Product documentation at CAMARA.
+  url: https://github.com/camaraproject/EdgeCloud
+servers:
+  - url: "{apiRoot}/{basePath}"
+    variables:
+      apiRoot:
+        default: https://localhost:443
+        description: API root.
+      basePath:
+        default: application-endpoint-discovery/vwip
+        description: Base path for the Application Endpoint Discovery.
+tags:
+  - name: Discovery
+    description: |
+      Find the closest App Instance Endpoint to the user device.
+paths:
+  /apps-endpoints:
+    get:
+      security:
+        - openId:
+            - application-endpoint-discovery:apps-endpoints:read
+      operationId: getAppEndpoints
+      parameters:
+        - name: filter
+          in: query
+          required: true
+          description: |
+            Filter the Edge Cloud Zones according to the parameter value.
+            For this API the only supported value is `closest`.
+          schema:
+            type: string
+            enum:
+              - closest
+        - name: IPv4-Address
+          in: header
+          required: false
+          description: The public IPv4 address allocated to the device.
+          example: "84.125.93.10"
+          schema:
+            $ref: "#/components/schemas/SingleIpv4Addr"
+        - name: IPv6-Address
+          in: header
+          required: false
+          description: The public IPv6 address allocated to the device.
+          example: "2001:db8:85a3:8d3:1319:8a2e:370:7348"
+          schema:
+            $ref: "#/components/schemas/DeviceIpv6Address"
+        - name: Network-Access-Identifier
+          in: header
+          description: |
+            3GPP network access identifier for the subscription
+            being used by the device.
+          example: "4d596ac1-7822-4927-a3c5-d72e1f922c94@domain.com"
+          required: false
+          schema:
+            $ref: "#/components/schemas/NetworkAccessIdentifier"
+        - name: Phone-Number
+          in: header
+          example: "+441234567890"
+          required: true
+          description: |
+            MSISDN in E.164 format (starting with country code) of
+            the mobile subscription being used by the device. Optionally
+            prefixed with '+'.
+          schema:
+            $ref: "#/components/schemas/PhoneNumber"
+        - name: AppId
+          in: header
+          required: true
+          description: |
+            A globally unique identifier associated with the application.
+            OP generates this identifier when the application is submitted over
+            NBI.
+          schema:
+            $ref: "#/components/schemas/AppId"
+        - name: X-Correlator
+          in: header
+          required: false
+          description: |
+            When the API Consumer includes the "X-Correlator" header in the
+            request, the API provider must include it in the response with
+            the same value that was used in the request. Otherwise, it is
+            optional to include the "X-Correlator" header in the response with
+             any valid value. Recommendation is to use UUID for values.
+          schema:
+            type: string
+            format: uuid
+      responses:
+        "200":
+          description: |
+            Successful response, returning the closest App Instance Endpoint
+            Zone to the user device identified in the request header.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AccessEndpoint"
+        "400":
+          $ref: "#/components/responses/400BadRequest"
+        "401":
+          $ref: "#/components/responses/401Unauthorized"
+        "403":
+          $ref: "#/components/responses/403Forbidden"
+        "404":
+          $ref: "#/components/responses/404NotFound"
+        "405":
+          $ref: "#/components/responses/405MethodNotAllowed"
+        "406":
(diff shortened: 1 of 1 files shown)


code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+  - url: "{apiRoot}/{basePath}"
+    variables:
+      apiRoot:
+        default: https://localhost:443
+        description: API root.
+      basePath:
+        default: application-endpoint-discovery/vwip
+        description: Base path for the Application Endpoint Discovery.
+tags:
+  - name: Discovery
```
gunjald, line 130: Should we name it something like Application Endpoint Discovery or similar?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+          in: query
+          required: true
+          description: |
+            Filter the Edge Cloud Zones according to the parameter value.
+            For this API the only supported value is `closest`.
+          schema:
+            type: string
+            enum:
+              - closest
+        - name: IPv4-Address
```
gunjald, line 151: Need to specify what all parameters combinations are valid as there are multiple header parameters proposed though it has been with other APIs too. IPv4-Address, IPv6-Address, Network-Access-Identifier and Phone-Number which is mandatory. Can all of them be present? If not then there should be some guidance for application developers what should be valid parameter combinations.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+        type: string
+        format: uuid
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier associated with the application.
+        OP generates this identifier when the application is submitted over NBI.
+    AccessEndpoint:
+      type: object
```
gunjald, line 270: I think for AccessEndpoint we need more discussion in terms of the data model. The reason is what should we expect in terms of the number of exposure endpoints of an application towards the application clients. May be an application is composed of multiple components and offers one or more endpoints to its clients for different purposes. In that case data model should consider those possibilities and accordingly consider them in defining the  AccessEndpoint structure e.g. one or more ports, fqdns etc as an example.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    * Connect their application client directly to the chosen application
+    endpoint, enhancing the responsiveness and reliability of the service.
+    * Deploy new instances if no suitable endpoints are within the identified
+    Edge Cloud Zone, optimizing their service distribution based on real-time
+    network geography.
+    # Quick Start
+    This API can be utilized by both client applications on user devices and
+    by servers. It uses a straightforward endpoint:
+    `/apps-endpoints?filter=closest`.
+    API consumers are required to authenticate via OAuth2 to obtain access
```
Kevsy, line 36: Suggest to remove the point about AuthN/AuthZ here because it is covered below in the dedicated section.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+        description: API root.
+      basePath:
+        default: application-endpoint-discovery/vwip
+        description: Base path for the Application Endpoint Discovery.
+tags:
+  - name: Discovery
+    description: |
+      Find the closest App Instance Endpoint to the user device.
+paths:
+  /apps-endpoints:
```
Kevsy, line 134: `apps-endpoints` or `app-endpoints`? I think it should be app-endpoints, because it is all the endpoints for a single application.. In fact `app-server-endpoints` may be best because we are only returning the server endpoint, not the client endpoint.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
Kevsy: Agreed, as you say it is the same for other APIs. I suggest we wait for Commonalities release 0.4 for guidance.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+      required: false
+      schema:
+        type: string
+        format: uuid
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier associated with the application.
+        OP generates this identifier when the application is submitted over NBI.
```
Kevsy, line 268: For the CAMARA API we should not say 'OP' or 'NBI', top avoid 'telco-specific language' . Can say 'This GUID will have been previously provided by the Edge Cloud Zone provider'. But how is that done? We should provide guidance...

code/API_definitions/Application_Endpoint_Discovery_API.yaml
Kevsy: Why `AccessEndpoint` and not `ApplicationServerEndpoint`?

Kevsy (commented):
Please see comments inline for review.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
gunjald: I agree to not to use OP.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+info:
+  title: Application Endpoint Discovery API
+  version: 0.1.0-wip
+  description: |
+    # Connect to the Nearest Application Endpoint
+    ---
+    # Overview
+    The Application Discovery API extends beyond the capabilities of the
+    Simple Edge Discovery API by not only locating the nearest Edge Cloud Zone
+    but also directly linking to the application endpoints within those zones.
```
gunjald, line 12: I think we should replace the zones with Edge Cloud Zones as it is the formal definition we have.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    This crucial enhancement enables immediate connection to the most relevant
```
gunjald, line 13: Should we keep words like "crucial enhancements"? Or we just write about the API intent?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    service endpoint, optimizing the application's interaction with
+    edge computing resources.
+    # Introduction
+    Operators often host multiple Edge Cloud Zones across various territories.
+    Connecting an application directly to the nearest operational service
```
gunjald, line 18: Do we have a concept for "nearest operational service"? Also, the application word should be qualified to not let the API consumers confuse it with edge application in my opinion.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    endpoint minimizes latency because data travels the shortest possible route.
```
gunjald, line 19: On this "minimizes latency because data travels the shortest possible route", I think we may say "it may minimize latency" but we should not say "shortest possible route" as it is subjective and this API may not be providing any guarantee for this.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    Unlike the Simple Edge Discovery API, which only provides the zone location,
+    this API calculates and provides the actual endpoints within the nearest
```
gunjald, line 21: May be we can say "this API provides the actual endpoints" i.e. drop the word "calculates".

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    zone. This approach ensures that developers can instantly integrate these
+    endpoints into their applications, improving performance and user
+    experience.
```
gunjald, line 24: Not sure if we should keep the word "performance" in "improving performance and user experience"?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    After identifying the nearest application endpoint through the API,
+    developers can:
+    * Connect their application client directly to the chosen application
+    endpoint, enhancing the responsiveness and reliability of the service.
```
gunjald, line 28: Should we keep "reliability"? Not sure if the API is suppose to provide any such capability?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+    * Deploy new instances if no suitable endpoints are within the identified
```
gunjald, line 29: In my opinion the API should not "Deploy new instances" if they are not available. The API should be explicit in what it intends to offer to its customer and that is defined as "Discovery" but not the instantiation in my opinion. Either there should be attribute/parameter which user can use to specify what to do if there is no endpoint available or API should return error.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+      IP-Address: 84.125.93.10
+    ```
+    Example 3: Request with Phone-Number
+    ```
+      GET /apps-endpoints?filter=closest HTTP/1.1
+      Host: example.com
+      Authorization: Bearer {your_access_token}
+      Accept: application/json
+      Phone-Number: +441234567890
+    ```
```
gunjald, line 81: I think one more example should be provided here for the "initial requests may be made without explicit device identifiers"

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+          required: true
+          description: |
+            Filter the Edge Cloud Zones according to the parameter value.
+            For this API the only supported value is `closest`.
+          schema:
+            type: string
+            enum:
+              - closest
+        - name: IPv4-Address
+          in: header
```
gunjald, line 152: I think we should mention about what are valid combination from the give set of the user identifiers e.g. IPv4-Address, Network-Access-Identifier etc. If one of them or any of them or all should be present?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: We don´t have an specific concept for it, but we have decided to rename it to "optimal Edge Cloud Zone" in order to avoid complexity

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree with you. We have added to the error 404 another option: "`EndpointNotFound`" . We agree it should return error. On the other hand, when a developer is aware that there is not any endpoints available (404), they can use the EdgeApplicationManagement to instantiate a new instance for that user.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
gunjald: To me  app-endpoints looks fine as you rightly said it is about the endpoints of the same application. The other one i.e. app-server-endpoints contains "server" keyword which i feel is more concrete and as we are referring mostly the "application" in general so app-endpoints looks more aligned. Any thoughts?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+            format: uuid
+      responses:
+        "200":
+          description: |
+            Successful response, returning the closest App Instance Endpoint
+            Zone to the user device identified in the request header.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AccessEndpoint"
```
gunjald, line 213: As the path itself says /apps-endpoint or if changed to /app-endpoints then shouldnt the AccessEndpoint be also made plural to indicate there could be more than one items depending upon the application design? We should keep options open for application with more than one endpoints while it allows to serve apps with single endpoint too.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: OK

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: OK. Agree with you. We are going to wait for the new commonalities to come out for guidance.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,596 @@
...
+      schema:
+        type: string
+        format: uuid
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier associated with the application.
+        OP generates this identifier when the application is submitted over NBI.
+    AccessEndpoint:
```
gunjald, line 269: As commented earlier an application may have more than one endpoints and for that reason we may use AccessEndpoints to accommodate plurality. If so, then we would need to make the AccessEndpoints as an array of objects that is being defined below which contains the endpoint accessibility information?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: OK. We are going to wait for commonalities 0.4.0 to come out for guidance.

code/API_definitions/Application_Endpoint_Discovery_API.yaml
gunjald: Also, in my opinion if multiple endpoints are returned then there should be an additional field which allow to embed a purpose or qualifier for a given endpoint which can be used by the client applications to differentiate one endpoint with other e.g. low quality video stream or high quality video stream endpoint? But then these qualifier should come from the API invoker may be at application onboarding time. Any thoughts?

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: Agree

code/API_definitions/Application_Endpoint_Discovery_API.yaml
LauraM-TEF: OK. Agree with AppInstanceEndpoint and as in the comment above in line 134, AppServerEndpoint is more specific and we are refering mostly the application in general

code/API_definitions/Application_Endpoint_Discovery_API.yaml
```diff
@@ -0,0 +1,609 @@
...
+  schemas:
+    AppId:
+      type: string
+      format: uuid
+      description: A globally unique identifier associated
+        with the application.
+        This GUID will have been previously provided by the
+        Edge Cloud Zone provider.
+    AccessEndpoints:
+      type: object
```
gunjald, line 272: I think one comment is lost for some reasons. From the last comment "May be an application is composed of multiple components and offers one or more endpoints to its clients for different purposes. In that case data model should consider those possibilities and accordingly consider them in defining the AccessEndpoint structure e.g. one or more ports, fqdns etc as an example."

In my opinion the AccessEndpoint should be an array of objects instead of a single object as it is defined now to have one or more endpoints. And each endpoint may have a qualifier defined by API invoker to help differentiate between the endpoints based on the purpose of the endpoint in client side code. May be such qualifiers can be provided during app onboarding time of the application along with the components to which the endpoint belongs to in the /app-endpoints response.

LauraM-TEF: Agree with you and as it was concluded in the edge cloud meeting for this version we keep a single endpoint

PR #252, opened 2024-06-06
EAM API: Add missing AppProvider

### LINKED ISSUE #251: Add Missing AppProvider to EAM
**Problem description**
AppProvider is missing from the Edge-Application-Management APIs.

**Expected behavior**
Expected that Apps and App Instances would have an AppProvider field to allow for multiple tenants. This is basically table stakes for any real platform implementation. It's unlikely an Operator would use a platform without being able to provide isolation between different users applications.

**Alternative solution**
None.

**Additional context**
AppProvider is already defined as part of the GSMA Federation EWBI API.

gainsley:
#### What type of PR is this?
* correction

Adds the missing AppProvider to the AppManifest model.

Fixes #251 

#### Changelog input

```
Add AppProvider field to AppManifest
```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -859,6 +859,7 @@ components:
       required:
         - name
         - version
+        - appProvider
         - packageType
         - appRepo
         - requiredResources
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -705,6 +705,8 @@ components:
           type: string
           pattern: ^[A-Za-z][A-Za-z0-9_]{1,63}$
           description: Name of the application.
+        appProvider:
+          $ref: '#/components/schemas/AppProvider'
         version:
           type: integer
           description: Application version information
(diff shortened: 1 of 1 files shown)


javierlozallu:
@gainsley, for this API an ID parameter to identify an App Provider may not be useful as the API implements 3-Legs Token so the App Provider will be known by this method.

gainsley:
Hi @javierlozallu, I'm not super-clear on what information will be in the authentication token, but typically a user can have permissions for multiple app providers (or all app providers, in the case of an admin). I would expect the auth token may provide for which data the user is allowed to access, but not which specific data (or AppProvider in this case) is being used in the context of the current API call. So I think the AppProvider field is still needed in the API call. Also tokens are typically valid for some time and can be used for multiple API calls before they expire, and the user may want to use different AppProviders for different API calls if they have permissions to do so, all while using the same auth token.

ThomasEdgeXR:
@javierlozallu i think we still need to solve this - as Jon said the authentication does not solve for knowing the AppProvider: as defined in https://github.com/camaraproject/EdgeCloud/blob/main/documentation/SupportingDocuments/Edge%20terminology/edge_terminology.md the AppProvider is the entity owning the application backend (i.e. the "developer organization" or "ISV" if you will). So it's not an individual account but a logical entity to which multiple users can belong. It is necessary to add this to the model so it is clear to which AppProvider this belongs. A workload belongs to one AppProvider, but one AppProvider may be linked with multiple user accounts.

crissancas (dismissed):

JoseMConde (approved):

Issue #277, opened 2024-06-26
TI API: 3Legs link obsolete


FabrizioMoggio:
**Problem description**
In the Documentation section, a link to Identity and Consent Management is not correct anymore: https://github.com/camaraproject//IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access/-and-user-consent.md

**Expected behavior**
a link to: 

https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md

is expected as described in the guidelines:(https://github.com/jpengar/Commonalities/blob/d4d44f5b72675d10192bc8fdceb21b299f576ec9/documentation/API-design-guidelines.md)

FabrizioMoggio:
Actually the link was just broken:

https://github.com/camaraproject//IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access/-and-user-consent.md

so we can keep the link just updating it as:

https://github.com/camaraproject/IdentityAndConsentManagement/blob/main//documentation/CAMARA-API-access-and-user-consent.md

FabrizioMoggio:
Fixed with: https://github.com/camaraproject/EdgeCloud/pull/278

FabrizioMoggio:
Merged

Issue #272, opened 2024-06-20
TI API: versioning


FabrizioMoggio:
**Problem description**
the "version" filed is not correct.

**Expected action**
fix it removing the release number

**Additional context**
as discussed for the CFS API here: https://github.com/camaraproject/CallForwardingSignal/discussions/27#discussioncomment-9701045

FabrizioMoggio:
fixed with: https://github.com/camaraproject/EdgeCloud/pull/278

FabrizioMoggio:
Merged

PR #278, opened 2024-06-26
TI API Rel 0.9.6: sourceTrafficFilters

FabrizioMoggio:
- removed rel number in "version" field according to:https://github.com/camaraproject/EdgeCloud/issues/272
- sourceTrafficFilters added according to: https://github.com/camaraproject/EdgeCloud/discussions/230#discussioncomment-9879076
- updated link to Identity and Consent: CAMARA-Security-Interoperability: https://github.com/camaraproject/EdgeCloud/issues/277
- Alignement with Device according to: https://github.com/camaraproject/EdgeCloud/discussions/247
- Alignement with "Edge Application Management API" according to: https://github.com/camaraproject/EdgeCloud/pull/278#discussion_r1666884747 and https://github.com/camaraproject/EdgeCloud/pull/278#discussion_r1666889871


FabrizioMoggio:
to be aligned with: https://github.com/camaraproject/EdgeCloud/discussions/247

code/API_definitions/Traffic_Influence.yaml
```diff
@@ -148,13 +148,15 @@ info:
     traffic toward a specific Application instance.\
     \
     **sourceTrafficFilters:**
-    The traffic can be from a specific application port in the Device.\
+    The traffic can be from a specific port in the device. If this parameter is
```
gunjald, line 151: I think we should mention the nature of the port which at other places said to be publicPort. Isn't it?

code/API_definitions/Traffic_Influence.yaml
```diff
@@ -166,7 +168,10 @@ info:
...
     different "trafficInfluenceID"). All the created resources are aggregated by
     the Application (identified by "appId"). The routing toward the selected
-    Application instance is only applied for provided user Devices.\
+    Application instance is only applied for provided user Devices. "publicPort"
+    can be used to identify the device. "publicPort" can be also used to
+    identify the flow to be influenced. If the flow to be influenced is from a
+    different port, "sourceTrafficFilters" can be used.\
     \
     **Notification URL and token:**
     Developers have a chance to specify call back URL on which notifications
```
gunjald, line 177: A minor one to make "call back" to callback i.e. to remove the whitespace.

code/API_definitions/Traffic_Influence.yaml
```diff
@@ -655,15 +658,15 @@ components:
             - 'deletion in progress'
             - 'deleted'
         sourceTrafficFilters:
-          description: Ports used locally by the device for flows to which
-           the requested traffic influence should apply. If omitted, then the
-           traffic influence will apply to all flows between the device and the
-           specified application server address and ports.
+          description: Port used locally by the device for flows to which
```
gunjald, line 661: Similar to previous comment to make port to publicPort?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+info:
+  title: Traffic Influence API
+  version: wip
+  description: |
+    ## Overview
+    The reference scenario foresees a Service, composed by one or more Service
+    Producers deployed in different geographical locations on Edge Cloud Zones
+    (Edge datacentres of Telco Operator) or in Cloud. The Service Producer,
+    deployed at the Edge, is referred as Edge Application Server (EAS).\
+    An Edge Cloud Zone is a platform in the Telco Operator network, offering
```
gunjald, line 15: "An Edge Cloud Zone is a platform" seems to be proposing a definition here. But we already have defined it at other place and we may use it just An Edge Cloud Zone without further qualifying here in my view.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    network, compute and storage resources to developers. A developer can
+    deploy and run applications on an Edge Cloud Zone, meaning reduced latency
```
gunjald, line 17: Not sure if this much of Edge Cloud Zone functionality we should mention here as it is elaborated as part of App LCM and SED API.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    to end users that are nearby, as the network path is shorter. A network
+    operator's EdgeCloud may comprise multiple Edge Cloud Zones, each in a
+    discrete location to bring latency benefits to end users across a country.\
+    The operator can help developers know which of the Edge Cloud Zones will
+    bring the best performance benefit for a given end user and application\.
+    The Traffic Influence API (TI API) provides the fastest routing from the
+    user Device (e.g. a Smartphone) to the optimal EAS instance in a specific
+    geographical location, installed in an Edge Cloud Zone.\
```
gunjald, line 25: I think the API tries to provide the fastest routing via an EAS that may be optimum to serve user Device in specific geographical location but it itself may not be located in the same geographical location as the text is suggesting. If the observation is correct we may need to rephrase the statement a bit.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    maybe for a set of users. Getting the optimal routing can be used to improve
+    latency maybe in combination with other CAMARA APIs such as QoD (Quality On
+    Demand). Providing the optimal routing is indeed an important step to get
+    the optimal latency.\
+    If the TI API is used to get the best routing at the Edge for a Device in a
+    geographical location and the Device moves to another geographical location,
+    the TI API can be invoked to get the optimal routing in the new geographical
+    location for that Device.
+    ## Introduction
+    The TI API provides the capability to establish the best routing, in terms
```
gunjald, line 37: I think we have changed the best to optimal for other APIs so may be here also we can use optimal in place of best.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    of latency, in a specific geographical area, between the user Device, e.g.
+    the user’s smartphone, and the optimal EAS instance nearby. If the Device
+    moves in a different geographical location where a more suitable EAS
+    instance is available, the TI API can be invoked to influence the Device
+    connectivity to get the optimal routing to the that local instance. It is i
+    mportant to notice that it is a task of the Application invoking the TI API
+    to detect the changes in the Device location.\
+    The generic architecture for the Service can foresee some Cloud instances of
+    the Application, one or more Edge Instances of the Application. A component
```
gunjald, line 46: Seems the word "component" occurred suddenly while not defined what it is referring to e.g. Cloud instances of the Application or Edge Instances- or both?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    of the Service is the TI API Consumer. This logical component can be
+    integrated in other components of the Service to invoke the TI API, creating
+    a "TrafficInfluence" resource specifying the desired intent.\
+    The TI API Producer implements the intent specified in the
+    "TrafficInfluence" resource.\
+    While the TI API can be invoked to activate the fastest routing for any
+    user, it can also be used to request the best routing for a specific user.
+    Invoking the TI API for each user, many "TrafficInfluence" resources are
+    created for each user to provide the requested routing for a set of users.\
```
gunjald, line 55: The statement "Invoking the TI API for each user, many "TrafficInfluence" resources are created for each user to provide the requested routing for a set of users." seems implicit as API in general is defined to target one user at a time so may omit it. 
Though for one user the TI resources could be multiple for a pair of source and destination port/protocol representing one traffic flow and which should be mention.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    The same approach is used for the geographical locations where the influence
+    of the traffic must be applied. Invoking the TI API without specifying a
+    geographical area activates the optimal routing toward any EAS instance,
```
gunjald, line 58: I think without specifying a geographical area the optimal routing should be towards a platform determined EAS instance which could possibly provide the optimal routing for a user in a given location rather than "any EAS"? Any thoughts?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    The TI API can be used to:
+    - optimise the routing when Devices need to consume the service provided
+        by a local EAS Instances.
+    - affect an already established Device routing when the Device moves
+        among different geographical locations. When the TI API consumer detects
+        a Device has entered a geographical location where an EAS instance is
+        available, it can invoke the TI API to get the optimal routing toward
+        that EAS instance.
+        If the Device moves to another geographical location, served by another
+        EAS instance, the routing might not be optimal anymore for the new EAS
```
gunjald, line 73: I think "the routing might not be optimal anymore for the new EAS" in this statement the "new EAS" is incorrect. It could be the existing serving EAS or current EAS which may not be suitable in new location of the user.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    provided by the Telco Operator and can also be used or retrieved with other
+    CAMARA APIs (“MEC Exposure & Experience Management API” and “Simple Edge
+    Discovery”). To add more regions the TI API must be invoked (POST) for each
+    region. New "TrafficInfluence" resources are created (with different
+    "trafficInfluenceID"). All the created resources are aggregated by the
+    Application (identified by "appId").\
+    \
+    **zone:**
+    The developer can specify in which geographical area he requires the fastest
+    routing toward the nearest Application instance. A "zone" is a smaller
```
gunjald, line 130: we should tell about the usage of zone and region and how an API user can figure out in which scenario he should be using which one. Also, somewhere we will need to then specify if a user can determine the relationship between zone and region say by another query API which may not be part of the TI API. But without knowing when region to be used and when zone is to be used may leave a gap in API consumer mind.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    **appId:**
+    A globally unique identifier associated with the application. This
+    identifier is provided during the application onboarding process.
+    To influence the traffic toward a specific Application. It is the Operator
+    Platform that detects the appropriate Application instance in the selected
+    "region" or "zone".\
+    \
+    **appInstanceId:**
+    A globally unique identifier generated by the Operator Platform to identify
+    a specific instance of the Application on a specific zone. To influence a
```
gunjald, line 147: Here is another related aspect of zone and region. The App instance id is referred to be associated to a zone but then if someone want to use region then how the instance id can be determined for the region?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    traffic toward a specific Application instance.\
+    \
+    **sourceTrafficFilters:**
+    The traffic can be from a specific port in the device. If this parameter is
+    used, the influenced flow is from the port defined in "sourceTrafficFilters"
```
gunjald, line 152: I suppose port is publicPort? isn't it?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    \
+    **destinationTrafficFilters:**
+    The Application can expose different service on different interfaces,
+    identified by port and protocol, with this parameter it is possible to
+    route the traffic toward a specific port and protocol exposed by the
+    Application.\
+    \
+    **Device:**
+    An user Device can be provided as an input. The Device can be identified by
+    the phone number (phoneNumber), an external identifier
```
gunjald, line 163: Though we may keep external identifier in the API but should we keep the explanation too as I understood from the commonality perspective that this identifier may be deprecated later?

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    (networkAccessIdentifier) or by its IP Address (Ipv4Address, Ipv6Address)
+    also specifying a Port. For IP address both the private (as seen from inside
+    the Device) and the public (as seen over Internet by a server connected to
+    the Device) can be used. To add more users the TI API must be invoked (POST)
+    of each user Device. New "TrafficInfluence" resources are created (with
```
gunjald, line 168: I think this is implicit "To add more users the TI API must be invoked (POST) of each user Device" and true of other APIs too. We may not need to write it explicitly. But just a suggestion.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    different "trafficInfluenceID"). All the created resources are aggregated by
+    the Application (identified by "appId"). The routing toward the selected
```
gunjald, line 170: This is also redundant in my opinion "All the created resources are aggregated by the Application (identified by "appId")" as it is not pointing to any functionality that the API user should be taking care or should keep in mind while invoking TI API.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    Application instance is only applied for provided user Devices. "publicPort"
+    can be used to identify the device. "publicPort" can be also used to
+    identify the flow to be influenced. If the flow to be influenced is from a
+    different port, "sourceTrafficFilters" can be used.\
+    \
+    **Notification URL and token:**
+    Developers have a chance to specify call back URL on which notifications
+    (e.g. session termination) regarding the session can be received from the
```
gunjald, line 178: I think so far nowhere it is mentioned the concept of "session" above. So session termination may need to be clarified. May be we should mention in above that the TI API creates a session with a given duration in which events can also be reported to the API consumers and session can be terminated by say DELETE or by the API server too.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    of users, the TI API must be invoked for each user creating a new
+    TrafficInfluce resource for each one.
+    If the Application offers different services on different interfaces a
+    traffic filter based on IP, Port and Protocol can be used. I this way it is
+    also possible to redirect different users to different interfaces.
+    Here are some possible intents:
+    1) activate the optimal routing for any EAS instance: the TI API must be
+    invoked with the "appId". The Telco Operator Platform identifies all the EAS
+    instances and activates the optimal routing on the Mobile Network.
+    2) activate the optimal routing in a specific Region or Zone: the TI API
```
gunjald, line 220: Like above comment i think a reference to Device should be included along with EAS.

code/Test_definitions/Traffic_Influence.yaml
```diff
@@ -0,0 +1,1030 @@
...
+    must be invoked with the "appId" and the Zones and Regions identifiers.
+    3) activate the optimal routing for a user devices: the TI API can  be
+    invoked with a user Device identifier (“Device”). For each user Device,
+    a TI API invocation is required.
+    ## Release Notes
+    The Traffic Influence API reduces the complexity of the 3GPP Traffic
```
gunjald, line 226: In my opinion we should not include a reference to 3GPP TI API as there could be different ways how this TI API can be implemented in a network and it is not intended to simply the 3GPP TI API.

FabrizioMoggio:
@gunjald I'm so sorry you did a great effort to go through the documentation. unfortunately that was the old version, I actually uploaded the new TI API in the wrong folder.

gunjald:
> @gunjald I'm so sorry you did a great effort to go through the documentation. unfortunately that was the old version, I actually uploaded the new TI API in the wrong folder. Anyway it was mostly the same file , just one line was different. so no problem

No worries :-) I hope we have better APIs and documents to help the adoptions.

code/Test_definitions/Traffic_Influence.yaml
gunjald: In Edge-Application-Management.yaml I see following definition:
__Edge Cloud Zone__ - An Edge Cloud Zone is the lowest level of
    abstraction exposed to an Application Provider who wants to deploy  an Application on Edge Cloud.
    Edge Cloud Zones exists within a Edge Cloud Region.

So the statement "An Edge Cloud Zone is a platform" is not same as defined in App LCM file. May be we can just use reference of App LCM definition here in TI.

code/Test_definitions/Traffic_Influence.yaml
gunjald: As mentioned in previous comment may be we can just refer to the definition of Edge Cloud Zone from Edge-Application-Management.yaml and not explain it here as there would be chance to get misalignment at two places.

FabrizioMoggio:
I also aligned the name of the parameters for Zones and Regions in the code.

- EdgeCloudRegion
- EdgeCloudZoneId

FabrizioMoggio:
@gunjald is there still pending on this PR? or can it be approved?

FabrizioMoggio:
@gunjald now all your comments should be addressed :-). Thank you very much for your support.

FabrizioMoggio:
Can we approve the PR?

nicolacdnll (approved):

Issue #256, opened 2024-06-07
EAM: Suggest that POST appinstance targets a single EdgeCloudZone


gainsley:
**Problem description**
Currently POST app instance targets multiple EdgeCloudZones, which creates some confusion as to what the correct behavior should be in some cases. The reason it is confusing is because we are effectively combining multiple API calls into one, where any single API call my behave differently from the other.

For example, suppose one of the EdgeCloudZones specified does not exist, but the other ones are ok. Does the API fail just one, or fail all? Is the return value 202 or 400?

**Expected behavior**
We should design APIs to be as simple as possible and do just one thing. The caller can manage calling the API multiple times if needed. So I suggest the POST app instance API target only a single EdgeCloudZone.

**Alternative solution**
Extensively document the expected behavior in various corner cases. Not really recommended.

**Additional context**

gainsley:
This is still an unresolved issue.

I suspect the intent was to allow deploying a large number of instances all at once. If that is still needed, then we need to change the response object to allow including both the successful AppInstanceInfo and an array of failure objects that includes the AppId, ZoneId, and error message. We may also want a mass-delete API, otherwise if the client can delete one-by-one, then the client can also certainly create one-by-one.

PR #280, opened 2024-06-27
Updating requiredResources in Application Management API

### LINKED ISSUE #253: EAM: Change RequiredResources numCPU to allow fractional cpus
**Problem description**
The Edge-Application-Management's current RequiredResources' numCPU field is an integer, which restricts vcpu resource allocation to whole vcpus. Notably Kubernetes allows for fractional vcpu allocation down to the millivcpu, to allow for very lightweight containers/microservices that do not use a lot of compute.

**Expected behavior**
Being able to specify required vcpus in milli-vcpu increments.

**Alternative solution**
None.

**Additional context**
The [GSMA Federation EWBI](https://www.gsma.com/solutions-and-impact/technologies/networks/gsma_resources/east-westbound-interface-apis-version-4-0/) already allows for this:
```
    Vcpu:
      type: string
      pattern: ^\d+((\.\d{1,3})|(m))?$
      description: Number of vcpus in whole, decimal up to millivcpu, or millivcpu format.
      example:
        whole:
          value: 2
        decimal:
          value: 0.500
        millivcpu:
          value: 500m
```
For VM-based deployments, fractional vcpus may not be allowed, but that should be implementation dependent. For example a VM deployed via Kata-containers (kube-virt) on a Kubernetes cluster may allow for fractional vcpus.

felipevicens:
**What type of PR is this?**

* enhancement

**What this PR does / why we need it:**

Enables the definition of different infrastructure for the application, 
allowing selection from among Kubernetes, Virtual Machines, and Containers. 
Additionally, the developer can specify the sizing of Kubernetes cluster for the 
application, the characteristics of compute resources, and the base add-ons 
to enabling monitoring and ingress in Kubernetes clustes.

Moreover, this PR changes `requiredResources` to a set of infrastructure
resorces, enabling developers to specify more than one resource for the
application. For example, an application composed of two containers
and one virtual machine.

**Which issue(s) this PR fixes:**

Fixes #253
Discussion #220

**Special notes for reviewers:**

```
This PR only applies to requiredResources in Application Management API
```

**Changelog input**

```
 Modify requiredResources in Application Management API
```


code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
...
+        version:
+          type: string
+          description: Minimum Kubernetes Version.
+          example: "1.29"
+        controlNodes:
+          type: integer
+          description: Number of nodes for Kubernetes control plane.
+          enum:
+          - 1
+          - 3
```
gainsley, line 1120: Not sure why an enum is defined for the controlNodes integer, shouldn't it allow any integer greater than 0?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
...
           description: Number of GPUs
 
+    Flavor:
+      type: string
+      description: |
+        Preset configuration for compute, memory, GPU,
+        and storage capacity. (i.e - A1.2C4M.GPU8G, A1.2C4M.GPU16G, A1.4C8M,..)
+      example: A1.2C2M.GPU8G
+
+    NodePools:
```
gainsley, line 971: Should be `NodePool`, not NodePools

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
...
+          description: Type of infrastructure for the application.
+          type: string
+          example: kubernetes
+          enum:
+            - kubernetes
+        version:
+          type: string
+          description: Minimum Kubernetes Version.
+          example: "1.29"
+        controlNodes:
```
gainsley, line 1115: Seems like there should be a `controlNodesFlavor` to indicate which flavor to use for control nodes? Or is it expected that the platform shall choose an appropriate size?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
           type: integer
           description: Number of GPUs
 
+    Flavor:
+      type: string
+      description: |
+        Preset configuration for compute, memory, GPU,
+        and storage capacity. (i.e - A1.2C4M.GPU8G, A1.2C4M.GPU16G, A1.4C8M,..)
+      example: A1.2C2M.GPU8G
```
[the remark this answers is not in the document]
javierlozallu: Agree with @gainsley, we decided in #220 not to use flavours but if it is a better solution to implement it, GET /edge-cloud-zones should return in the response the information about the available flavour for each edge-cloud-zone of interest.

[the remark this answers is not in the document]
gunjald: I also tends to agree :-) May be with GET /edge-cloud-zones we can add query parameters to retrieve list of flavors and then in future we can also extend other resources via query parameters. Just a suggestion though.

code/API_definitions/Edge-Application-Management.yaml
gainsley: Sorry, I realize now this is for master nodes, so ignore my previous comment. [Docs](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/#multi-node-etcd-cluster) recommend up to 5 nodes for large clusters, I don't think we'll be dealing with large clusters here but perhaps for completeness add an enum value for 5.

code/API_definitions/Edge-Application-Management.yaml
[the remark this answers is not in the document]
felipevicens: Agree, I'll add an entry in GET /edge-cloud-zones to report the flavors.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
...
+      description: |
+        Addons for the Virtual Machine.
+      type: object
+      properties:
+        dockerCompose:
+          type: boolean
+          example: true
+          default: false
+          description: |
+            Enable docker-compose in the virtual machine to deploy applications.
```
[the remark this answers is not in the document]
felipevicens: Sounds good, I'll remove the VM addon and create a docker-compose type.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -1024,33 +1249,23 @@ components:
...
-          description: Storage in giga bytes
-        gpu:
-          type: array
-          description: Number of GPUs
-          items:
-            $ref: '#/components/schemas/GpuInfo'
+      type: array
+      items:
+        oneOf:
+        - $ref: "#/components/schemas/Kubernetes"
```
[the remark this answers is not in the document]
felipevicens: The idea is to manage applications as self-contained units, including the resources they need. This way, the application itself collects all the resources required to work properly. If a second application needs to be deployed on the same cluster, it might indicate that the resource requirements for the first application were overestimated. Here are two options for the developer:

a) Modify the Helm chart to add the application there (even modify the resources to fit - this would require an API for application Update) or
b) Deploy a new application.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -1024,33 +1249,23 @@ components:
...
+        - $ref: "#/components/schemas/VirtualMachine"
+        - $ref: "#/components/schemas/Container"
```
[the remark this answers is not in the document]
felipevicens: Right!, I'll modify it.

code/API_definitions/Edge-Application-Management.yaml
felipevicens: Okey, the idea here was to offer Control Plane w/HA and wo/HA but at the end it is controlled by the operator. I think better approach is a boolean controlPlaneHa: true/false

code/API_definitions/Edge-Application-Management.yaml
felipevicens: Good catch! changed

code/API_definitions/Edge-Application-Management.yaml
felipevicens: The definition of control nodes is out of the scope for the Application Developer.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +961,231 @@ components:
...
 
+    Flavor:
+      type: string
+      description: |
+        Preset configuration for compute, memory, GPU,
+        and storage capacity. (i.e - A1.2C4M.GPU8G, A1.2C4M.GPU16G, A1.4C8M,..)
+      example: A1.2C2M.GPU8G
+
+    NodePools:
+      description: |
```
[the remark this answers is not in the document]
gunjald: This looks to be very resource heavy approach by one app to one type of infra like a k8s cluster unless we provide a way to enable in some way deploy multiple applications on one cluster. Also what will happen if cluster creation fails? That means application onboarding failed as both are now one atomic package. And another issue could be once an app along with its given infra accepted I cannot change the infra e.g. reduce or increase the resources if needed.

So I still think specially with cluster type of infra that it will be hard to implement which could mean creating a cluster dynamically which could be a very time consuming process. If we delink infra creation then there could be options like platform offline creates cluster and provide API to retrieve details of cluster ID or even provide infra creation API to manage infra for applications and use the information with the App LCM API to link them together.

Means there could be ways but otherwise in terms of approach it seems to be tightly couple the infra and applications and may reduce reusability. May be more inputs will help here.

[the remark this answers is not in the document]
felipevicens: Hi @gunjald,

Sounds good. I think it would be interesting to discuss the creation of an API to manage the infrastructure lifecycle (Create, Update, Delete). Enabling the Kubernetes cluster reference within the Application Management API would be easy.

For now, I think it's safe to keep things this way, allowing developers to use a Kubernetes cluster and define the minimum configuration details required by their application. We can then open a discussion about how to design a more comprehensive API for infrastructure management resources.

code/API_definitions/Edge-Application-Management.yaml
nicolacdnll: I'm not sure if this should be part of the API. Application provider should have SLA in place with the operators and they shouldn't care about how the control plane of the operator infra is implemented.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -925,6 +931,14 @@ components:
...
       default: unknown
 
+    EdgeCloudZoneFlavors:
+      description: List of unique Name IDs of Infrastructure Flavors
+      type: array
+      items:
+        type: string
+        description: Flavor ID
+        example: A1.2C2M.GPU8G
+
```
nicolacdnll, line 941: Shouldn't each flavor have the list and spec of resources they provide?

felipevicens: I Agree Nicola, good point. I'll convert the string to object with the spec information. Thanks!

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -897,13 +897,19 @@ components:
         and the value of the Edge Cloud Provider
         object. This value is used to identify an Edge Cloud zone
         between Edge Clouds from different Edge Cloud Providers.
+      required:
+        - edgeCloudZoneId
+        - edgeCloudZoneName
```
gunjald, line 902: How is uniqueness defined here? Is edgeCloudZoneId or edgeCloudZoneName are unique or their combination is?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -925,6 +931,38 @@ components:
         - unknown
       default: unknown
 
+    EdgeCloudZoneFlavors:
+      description: List of unique Name IDs of Infrastructure Flavors
```
gunjald, line 935: How should a flavor be visualized from a developer point of view. Is a flavor represent a virtual machine (VM) or a server node with given set of resources mapped to it?

gunjald: If say ir represent a single node then gpuMemory attribute alone may not be sufficient to allocate such a resource. There may be attributes like gpuCount, gpuFamily etc are also to be considered to meet the workload requirements which need GPU.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -925,6 +931,38 @@ components:
         - unknown
       default: unknown
 
+    EdgeCloudZoneFlavors:
```
gunjald, line 934: There is another parameter "Flavor" below in yaml. How does that correlates with EdgeCloudZoneFlavors?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +999,278 @@ components:
...
+        - numNodes
+      properties:
+        name:
+          type: string
+          example: nodepool1
+          description: |
+            Nodepool Name (Autogenerated if not provided in the request)
+        flavor:
+          $ref: '#/components/schemas/Flavor'
+        numNodes:
```
gunjald, line 1024: Shouldnt it be something like numFlavors for better correlation?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +999,278 @@ components:
...
+        - nodePools
+        - infraKind
+        - applicationResources
+      properties:
+        infraKind:
+          description: Type of infrastructure for the application.
+          type: string
+          example: kubernetes
+          enum:
+            - kubernetes
```
gunjald, line 1137: The infrakind is part of the top level attribute KubernetesResources and looks redundant with value as "kubernetes" as KubernetesResources itself indicate that it is kubernetes resource.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +999,278 @@ components:
...
+        millivcpu, or millivcpu (i.e 500m) format.
+      example: "500m"
+
+    KubernetesResources:
+      description: Definition of Kubernetes Cluster Infrastructure.
+      type: object
+      required:
+        - nodePools
+        - infraKind
+        - applicationResources
```
gunjald, line 1130: In my view we should keep applicationResources separate from KubernetesResources. KubernetesResources should point to k8s related information while applicationResources may be a consumer of KubernetesResources. But KubernetesResources should not contain applicationResources and create a tight coupling.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -961,6 +999,278 @@ components:
...
+        addons:
+          $ref: '#/components/schemas/K8sAddons'
+
+    KubernetesResourcesRef:
+      type: object
+      description: Kubernetes reference object
+      required:
+        - infraKind
+        - kubernetesClusterRef
+        - applicationResources
```
gunjald, line 1173: As my previous comment i suggest to delink applicationResources from KubernetesResourcesRef. We may keep KubernetesResourcesRef to associate kubernetes entities only and let other abstractions which needs it can refer to it.

code/API_definitions/Edge-Application-Management.yaml
Kevsy: SimpleEdgeDiscovery has:

  * `edgeCloudZoneId` is a UUID for the Edge Cloud Zone.
  * `edgeCloudZoneName` is the common name of the closest Edge Cloud Zone to
  the user device.
  * `edgeCloudProvider` is the name of the operator or cloud provider of
  the Edge Cloud Zone.
  
  So, `edgeCloudZoneId` is expected to be unique, e.g. a namespaced URN

code/API_definitions/Edge-Application-Management.yaml
felipevicens: Hi @gunjald, given that flavors introduce complexity to the API, I'll implement a change that removes them. This will provide greater flexibility for operators to allocate workloads of any size.

code/API_definitions/Edge-Application-Management.yaml
felipevicens: I agree, We will propose a change for that.

LuisVTazon: The latest commit to this PullRequest ontroduces modifications to cover this and previous comments. **The concept of flavours has been removed** so that the cpu/memory required for an application is directly defined by the API consumer (the application porvider). 

For ths case of kubernetes, the case **where an application is expected to be deployed over a k8s cluster**, the concept of _resource topology_ is  introduced in order to allow the API consumer (the application prvider) describing what is the minimum k8s infrastructure  required to host the application. The topology describes the minimum number of nodes required and the minimum cpu/memory required per node. 

**Also,** in order to cope with applications with different components that need different hardware (e.g.: GPU required in some component) the concept of hardware pool has been introduced to allow th application provider to define the specific hardware configuration for each application component. **With this information, the API server (i.e.: an edge orchestrator) will be able to determine the requirements of the k8s cluster and nodepools needed, to host the application components**. In this proposal we start with 2 types of pools (CPU and GPU) that can be later extended to additional specific hardware (e.g.: future TPU or other specific processors to consider in the future).

**Finally**, to cope with scenarios of shared k8s clusters, the parameters _isStandalone_ and _kubernetesClusterRef_ have been added.

code/API_definitions/Edge-Application-Management.yaml
gainsley: I would like to point out that if edgeCloudZoneName (or edgeCloudZoneName + edgeCloudProvider) cannot uniquely identify the zone, i.e. only the UUID can uniquely identify the zone, then we won't be able to support a declarative API. That's probably ok if the API is mainly going to be accessed via a GUI by a human, but if we want to support automation and infra-as-code via yaml files, it would be much nicer to be able to have a declarative API.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -908,6 +912,8 @@ components:
           $ref: '#/components/schemas/EdgeCloudProvider'
         edgeCloudRegion:
           $ref: '#/components/schemas/EdgeCloudRegion'
+        edgeCloudKubernetesClusterRef:
+          $ref: '#/components/schemas/KubernetesClusterRef'
```
gainsley, line 916: Sorry, just noticed this, but I don't think the kubernetes cluster ref belongs here. It's not really a property of an EdgeCloudZone. I think we need a separate object to be able to specify the optional cluster ref as part of the create appinstance API that can be arrayed. Also discussion #256 is relevant here.

felipevicens: Hi @gainsley, I've added for the GET /apps/{appId}/instances to retrieve the cluster where an application is deployed. I've noticed that it also impacts on the POST where it is only specified the edgeCloudZone. What do you think if we put the k8s cluster ref outsie from the edgeCloudZone but inside of appInstanceInfo?.

code/API_definitions/Edge-Application-Management.yaml
felipevicens: Would the edgeCloudZoneId be the only required parameter for the edgeCloudZone?

code/API_definitions/Edge-Application-Management.yaml
gainsley: Yes, I think it makes sense to have the clusterref in appInstanceInfo. Unfortunately though, that doesn't help the `post /app/{appId}/instances` case since that doesn't use `appInstanceInfo`. There I think you'll need to create a new wrapper object that is { clusterRef (optional), EdgeCloudZoneId } that can then be arrayed.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -355,7 +355,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/AppZones'
+              $ref: '#/components/schemas/AppInstaces'
```
gainsley, line 358: typo: AppInstaces -> AppInstances

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -683,16 +683,25 @@ components:
               accessPoints:
                 $ref: '#/components/schemas/AccessEndpoint'
           minItems: 1
+        kubernetesClusterRef:
+          $ref: '#/components/schemas/KubernetesClusterRef'
         edgeCloudZone:
           $ref: '#/components/schemas/EdgeCloudZone'
 
-    AppZones:
+    AppInstaces:
```
gainsley, line 691: typo: AppInstaces -> AppInstances

gainsley (approved):
Thanks for fixing that, looks good, just need to fix the typo.

gainsley:
Not for this PR, but we may want to consider adding a disk space requirement to the Kubernetes Topology and VmResources/ContainerResources/DockerComposeResources objects. I see they have `AdditionalStorage` fields, but that seems to be something in addition to the root partition. Users should be able to specify the requirements for the root partition size.

gainsley:
Also for a follow-up PR, for GPU-specific resources, it does make sense that users would want to be able to reserve GPU VRAM for their LLM/etc usage, but Kubernetes really only allocates gpu resources based on number of gpus. A better understanding of what [Kubernetes provides](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/) and what the [Nvidia operator provides](https://github.com/NVIDIA/k8s-device-plugin#quick-start) would inform how users would actually be able to specify GPU resources. I think we probably need a num GPUs field, and *maybe* a GPU type/family as well, but it needs some exploration.

felipevicens:
> Also for a follow-up PR, for GPU-specific resources, it does make sense that users would want to be able to reserve GPU VRAM for their LLM/etc usage, but Kubernetes really only allocates gpu resources based on number of gpus. A better understanding of what [Kubernetes provides](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/) and what the [Nvidia operator provides](https://github.com/NVIDIA/k8s-device-plugin#quick-start) would inform how users would actually be able to specify GPU resources. I think we probably need a num GPUs field, and _maybe_ a GPU type/family as well, but it needs some exploration.

Hey @gainsley,

When setting the required GPU RAM, I was thinking about using NVIDIA's MIG strategy (Multi-Instance GPU). (https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-mig.html#about-multi-instance-gpu | https://docs.nvidia.com/datacenter/tesla/mig-user-guide/#a100-mig-profiles)

MIG allows for hard partitioning of large GPUs, enabling workloads to be allocated to specific portions. This way, the operator can share and allocate the resources of a single big GPU across multiple workloads.

I agree, we need to explore this further. We'll likely need to add an array of GPUs, such as: [ {"memory": "4G", "family": "Hopper"}, {"memory": "24G", "family": "Ampere"}]

gainsley:
Thanks for fixing the typos. Yes let's discuss more how to specify gpus in a separate issue.

gainsley (dismissed):

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -683,16 +683,25 @@ components:
               accessPoints:
                 $ref: '#/components/schemas/AccessEndpoint'
           minItems: 1
+        kubernetesClusterRef:
+          $ref: '#/components/schemas/KubernetesClusterRef'
         edgeCloudZone:
           $ref: '#/components/schemas/EdgeCloudZone'
 
-    AppZones:
+    AppInstances:
```
gunjald, line 691: The name "AppInstances" seems to be conveying the intent that it refers to information about the instances of the application while the object itself and description says it is collection of edge cloud zone where the applications are to be deployed. It seems there is some anomaly here that may be corrected.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -683,16 +683,25 @@ components:
               accessPoints:
                 $ref: '#/components/schemas/AccessEndpoint'
           minItems: 1
+        kubernetesClusterRef:
```
gunjald, line 686: It is not clear how developer will get the KubernetesClusterRef? Are there suppose to be any query API to find it out or how do we see retrieving this information by the API user?

code/API_definitions/Edge-Application-Management.yaml
gainsley: I agree, I think it may make more sense to call it "AppZones" as it was before.

code/API_definitions/Edge-Application-Management.yaml
gainsley: There are two ways to get a ClusterRef:
1) From the result of a previous Create AppInstance. Creating an AppInstance will use (or create) an underlying cluster for Kubernetes/Helm apps. After the instance is created, querying the AppInstanceInfo will give the clusterRef of the cluster the instance was deployed to. The use case here is to allow the user to deploy multiple AppInstances to the same cluster (for example database, tracing, logging, workload, frontend etc sets of instances that can communicate with each other in the same cluster).
2) From a cluster management API. The system may provide a cluster management API outside of the NBI, or this EAM CAMARA NBI API may be extended in the future to provide a cluster management API. The clusterRefs can then come from a "GET clusters" API.

code/API_definitions/Edge-Application-Management.yaml
gainsley: The entire KubernetesResources object is intended to describe the Application's required resources for a Kubernetes deployment. So everything in it is in fact application resources. But having another "ApplicationResources" field inside KubernetesResources I think causes confusion. I would suggest simply flattening out that field, and including CpuPool and GpuPool directly into the KubernetesResources object. Or perhaps renaming ApplicationResources to NodePools.

JoseMConde (approved):

JoseMConde (approved):

Issue #291, opened 2024-09-09
TI API: resource modification while action is still in progress


FabrizioMoggio:
**Problem description**
- there is no error handling if the API Consumer invokes the TI API for the same Device-AppId if the previous requests are not completed yet.
- the modification of the resource is expected to be done via PATCH but this is not well define din the documentation

**Expected behavior**
- improve error handling considering to provide back an error code if the status of the previous action on that Device/AppId is not yet "active".
- improve documentation
- suggest to use Patch to modify a resource instead of creating a new one for the same Device-AppId

**Additional context**
If, the status of the first resource is "active" and the API Consumer, to modify the routing, instead of using a PATCH, creates a new resource with another POST, this is allowed and the new resource will overrule the previous one.

FabrizioMoggio:
PR merged

Issue #294, opened 2024-09-12
TI API: destination EAS IP address


FabrizioMoggio:
**Problem description**
The main exposed Intent is: "activate the optimal routing for the closest EAS instance". How to "point" to that Instance is not currently in scope. Usually that information is already known by the Application, because the EAS Instance was deployed via the Edge Cloud APIs but maybe, for some reason, also related to the network, the chosen EAS, in that very moment, could not be the one retrieved by the other EdgeCloud APIs. For this reason we should provide back the EAS IP address. This should solve any problem related to the Client Application targeting the right EAS Instance and it will be clear for the Developer what to expect.

**Possible evolution**
Add the EAS IP address in the TI API callback response to the TI API Consumer

**Alternative solution**
Just relay on the usage of FQDN with EDGE DNS providing the EAS Instance IP: https://github.com/camaraproject/EdgeCloud/issues/293

**Additional context**
https://github.com/camaraproject/EdgeCloud/issues/293#issuecomment-2345585435
https://github.com/camaraproject/EdgeCloud/issues/293

FabrizioMoggio:
Fixed with: https://github.com/camaraproject/EdgeCloud/pull/311

FabrizioMoggio:
PR merged

Issue #293, opened 2024-09-09
TI API: DNS cache


FabrizioMoggio:
**Problem description**
for the traffic flow, to be redirected to the Edge, if the application always use the same FQDN, it is a task of the Edge DNS to associate the FQDN to the IP address of the local EAS Instance. This mechanism doesn't work if the App Client caches the DNS resolution

**Expected action**
improve the TI API documentation

**Additional context**
https://github.com/camaraproject/EdgeCloud/discussions/292

gunjald:
I think in context of TI API we may need to talk about two scenarios. 1) When the Application Client on UE say an Android Application triggers a TI query, 2) When any other app say a cloud based application triggers the TI API for a given UE.

In 1), the Application Client source code has the logic that it has initiated a TI API and based on successful outcome it may expect the DNS query response will be updated with the target EAS instance though it is subjective that how long it can take the network to update. But from application client perspective it may need to keep trying sending the query till the received response contains the EAS IP which is different than the EAS IP it was earlier using.

In 2nd scenario, if the 3rd Party App fires the TI API for a UE I would expect that there is custom APIs which the Application client might be using to talk to 3rd party cloud Application and the cloud App will need to indicate the Application client to start the DNS query as the EAS IP may be changed due to TI API invocation and DNS cache will be updated soon. From this point onward the same behavior will be expected at Application Client as described in scenario 1 above.

Though this is my understanding that how TI API flow will work specially from the app client perspective but would like to know if this is correct or I am missing something? But one of the key point is that the network DNS is assumed to be updated as part of the TI API implementation to reflect the IP address of the EAS for application FQDN used by the app client on device.

FabrizioMoggio:
Both the scenarios seams correct to me.

About the DNS update, I don't see how this is a task of the TI API. 

The TI API leverages on the presence of an EAS instance up and running, with all the configuration needed, in the Edge Data Network. 

When the EAS is installed, its endpoints must be exposed and the DNS must be configured for it to be usable. This must be done when the application is deployed and activated. This action must be performed if the Application is exposed via Internet (Public DNS configuration) or at the Mobile Network Edge (Edge DNS configuration). If a DNS configuration procedure is not active on the "deployment and activation" workflow, the Application will be never reached (independently by the TI API).

Is not suck a task (DNS Configuration) foreseen in the Edge Cloud API for application deployment? 
Not part of the NBI but part of the SBI obviously.

gunjald:
[...]

PR #296, opened 2024-09-12
TI API: Resource modification

FabrizioMoggio:
Resource modification while action is still in progress.

#### What type of PR is this?
* enhancement/feature
* documentation

The PR improves the error handling when the API Consumer invokes the TI API for the same Device-AppId if the previous requests are not completed yet.

Resource modification while action is still in progress: https://github.com/camaraproject/EdgeCloud/issues/291

### DIFF
--- a/code/API_definitions/Traffic_Influence.yaml
+++ b/code/API_definitions/Traffic_Influence.yaml
@@ -92,6 +92,54 @@ info:
         Application detects a location change, it can invoke the TI API again to
         request a new routing optimization toward the new EAS instance.
     ## Quick Start
+    The TI API is consumed by an Application Function (AF) requesting for the
+    optimal routing, in term of latency, for the traffic flow from a Device
+    toward EAS instances in Edge Cloud Zones.\
+    When the Application (the EAS) is onboarded and deployed in the Edge Cloud
+    Zones, the Application is identified with a unique identifier ("appId").\
+    Using the application identifier ("appId") and specifying a Zone or a Region
+    the Telco Operator Platform, autonomously identifies the best instance of
+    the EAS toward which routing the traffic flow and configures the Mobile
+    Network accordingly to get the fastest routing.\
+    If just the application identifier is used, the Telco Operator Platform
+    identifies all the EAS Instances and activates the optimal routing on the
+    Mobile Network.\
+    If the optimal routing in term of latency should be available just for a set
+    of users, the TI API must be invoked for each user creating a new
+    TrafficInfluce resource for each one.
+    If the Application offers different services on different interfaces a
+    traffic filter based on IP, Port and Protocol can be used. I this way it is
+    also possible to redirect different users to different interfaces.
+    Here are some possible intents:
+    1) activate the optimal routing for the closest EAS instance: the TI API
+    must be invoked with the "appId". The Telco Operator Platform identifies
+    all the EAS instances and activates the optimal routing on the Mobile
+    Network so that the user is connect to the closest EAS insteance.
+    2) activate the optimal routing in a specific Region or Zone: the TI API
+    must be invoked with the "appId" and the Zones and Regions identifiers.
+    3) activate the optimal routing for a user Device: the TI API can  be
+    invoked with a user Device identifier (“Device”). For each user Device,
+    a TI API invocation is required. The optimal routing is provided only for
+    the requested Devices.    
+    ## Authentication and Authorization
+    CAMARA guidelines defines a set of authorization flows which can grant API
+    clients access to the API functionality, as outlined in the document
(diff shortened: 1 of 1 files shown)


FabrizioMoggio:
It is used:  GENERIC_409_DENIED_WAIT

code/API_definitions/Traffic_Influence.yaml
```diff
@@ -92,6 +92,54 @@ info:
...
+    identifies all the EAS Instances and activates the optimal routing on the
+    Mobile Network.\
+    If the optimal routing in term of latency should be available just for a set
+    of users, the TI API must be invoked for each user creating a new
+    TrafficInfluce resource for each one.
+    If the Application offers different services on different interfaces a
+    traffic filter based on IP, Port and Protocol can be used. I this way it is
+    also possible to redirect different users to different interfaces.
+    Here are some possible intents:
+    1) activate the optimal routing for the closest EAS instance: the TI API
```
gunjald, line 114: A question on 1) when we say "activate the optimal routing for the closest EAS instance".

Does it mean that if the TI API (step 1) is triggered by a UE application and after that it may need to wait for the TI resource state to become "active". Once the state is "active" the UE application can assume that the UE may been moved to the optimal UPF near to the given EAS serving that location which is the job of the TI API.
Now in this active state the UE application can find endpoint of the EAS instance (i.e. appId) by using the Application Endpoint discovery API (step 2) which I presume that will deliver the EAS instance which is optimal or closest to UE as the earlier call to TI API was successful and hence due to right UPF assignment the traffic will be automatically optimally routed to EAS.

The above explanation is in relation to ongoing discussion #293 and written to clarify following  points,
1) The UE can only send traffic to an EAS whose IP is available to it before starting any traffic towards the EAS
2) Once a packet has been generated with a given destination IP, this packet can only be processed by the host or an application (EAS) on that host
3) There may be multiple routes between a given set of source and destination address with different traits in terms of QoS

My understanding is that TI API is mainly concerned with point 3 where it can attempt or influence network to reselect a UPF for a UE which can optimally serve a given UE trying to access an EAS in its current location. 
Any traffic from UE via this UPF to an EAS which is hosted to an edge cloud accessible via this UPF is assumed to be on "optimal routed". 
Hence once the TI API is successful then the UE needs to determine the EAS IP via other ways to as explained above but it is not the TI API task to deliver the EAS endpoints to UE.

I may be repeating the statements but only objective is to understand what actions are to be done in my API client application and what is to be taken care in the TI API implementation.

If require I may prepare few slides to explain and get your views.

JoseMConde: Hi @gunjald @FabrizioMoggio  , If you want, Deepak could prepare few slides and I can reserve some slot on our meeting and disscus about this topic. Maybe tomorrow is a lit bit hasty, next meeting on october first, could fit for you??

FabrizioMoggio: According to my understanding of the "story" described by @gunjald, his conclusion is correct. 

- The optimal routing is active only when the TI API invokes the callback to inform the API Consumer about the "active" status
- From that moment the Service can instruct the App Client on which is the destination for the communication flow.
- That destination must be the same used in the TI API invocation. 
- with: https://github.com/camaraproject/EdgeCloud/issues/294. The task for the Service will be easier 
- When the App Client generates a flow toward that destination the network had already setup the optimal routing.

@JoseMConde for sure we can discuss about this in the next Edge Cloud call

One note: is this discussion preventing to approve this PR? The text we are commenting was already in the YAML, just moved to improve readability. The actual value of this PR is in the new error introduced (GENERIC_409_DENIED_WAIT) that prevents the modification of a requested intent while it is still under implementation. I suggest to approve this PR and we can open a new issue related to this discussion if required.

JoseMConde: Hi @FabrizioMoggio,
You are totally right, we can approve this PR and open a new issue with this topic. I also agree with you about the conclusions of the Deepak's explanation, but maybe It makes sense to open this issue to see what the rest of the group thinks about it.
Could you do it for me??
Thanks. :)

code/API_definitions/Traffic_Influence.yaml
```diff
@@ -1025,6 +967,46 @@ components:
...
+                message: The resource that a client tried to create already
+                 exists.
+            GENERIC_409_CONFLICT:
+              description: Duplication of an existing resource
+              value:
+                status: 409
+                code: CONFLICT
+                message: A specified resource duplicate entry found.
+            GENERIC_409_DENIED_WAIT:
+              description: Specific conflict situation that is relevant in
```
FabrizioMoggio, line 1000: This text must be updated and specialized for the TI API

crissancas (approved):

PR #302, opened 2024-09-18
EAM API: remove appid from appinstance path

### LINKED ISSUE #299: EAM: remove appId from get app instances API
**Problem description**

Currently the Edge Application Management API for retrieving app instances requires the client to specify the AppId. If the AppProvider has multiple different applications they care about, then the AppProvider needs to: 1) call the getapps API to get the IDs for their apps, 2) for each appId call the get appinstances API. This requires some logic client side and the overhead of multiple APIs calls. Or, if the client is really basic (auto-generated, etc), then the user themselves must run the getapps API, then copy the AppId, then paste it into the get appinstances API, and potentially do that multiple times.

It is very common for an AppProvider to have multiple applications, so this is a common scenario. Consider also that every client (web, cli (different languages)) will need to reimplement this same logic of getting the App IDs and looping over them.

**Possible evolution**

I suggest that we move the appId in the current `get /apps/{appId}/instances` path into the query parameters, so it is treated as an optional filter, much like `region` and `status` are filters for `get /edge-cloud-zones`. The new api path will then be `get /appinstances`. This will allow for a single API call to get all or a filtered set of App Instances.

For consistency, we will also need to:
- change `post /apps/{appId}/instances` into `post /appinstances`, moving the appId into the request body.
- change `delete /apps/{appId}/instances/{appInstanceId}` into `delete /appinstances/{appInstanceId}`. I believe this is actually more accurate because appInstanceId by itself uniquely identifies the app instance, and appId is not necessary.

**Alternative solution**

None.

**Additional context**

gainsley:
#### What type of PR is this?

* enhancement/feature

Removes appId from the appinstances path. For retrieval, it becomes an optional parameter, allowing a single API call to get appinstances from different Apps. For delete, it is not necessary because appInstanceId is sufficient to identify the instance.

Fixes #299 

#### Changelog input

```
 release-note

```

#### Additional documentation 

```
docs

```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -326,7 +326,7 @@ paths:
         '503':
           $ref: '#/components/responses/503'
 
-  /apps/{appId}/instances:
+  /appinstances:
     post:
       security:
         - openId:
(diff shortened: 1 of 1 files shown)


JoseMConde:
Hi @gainsley , one thing about this:
for the get and the delete method, looks good to me, but I have doubts about the post method, because if we make appId as optional how do we know which App is going to be instantiate??
If  we don't add the appId on the requested body, could we instantiate all the Apps by mistake,??

gainsley:
Hi @JoseMConde thank you for catching that, indeed my change made the appId and appZone optional which is wrong. I've added the required spec so that they are required. Thanks for reviewing!

JoseMConde (approved):

PR #304, opened 2024-09-19
Update CODEOWNERS

maheshc01:
#### What type of PR is this?

* subproject management

as discussed during the call adding myself as code owner.

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,7 +6,7 @@
 
 # These are the default owners for the whole content of this repository. The default owners are automatically added as reviewers when you open a pull request, unless different owners are specified in the file.
 
-* @gunjald @kevsy @crissancas @FabrizioMoggio @nicolacdnll @gainsley @javierlozallu
+* @gunjald @kevsy @crissancas @FabrizioMoggio @nicolacdnll @gainsley @javierlozallu @maheshc01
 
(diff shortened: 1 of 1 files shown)


eharrison24:
Hi Mahesh,

This might be a question that goes beyond just you, for the codeowners of the repos do we think it would be possible to input their first and last name, as well as their github id? Sometimes it's a little difficult to track down who the github id belongs to. This could be a relatively easy solution? @hdamker @wrathwolf thoughts on this idea? 

Best,
Evan

hdamker:
> This might be a question that goes beyond just you, for the codeowners of the repos do we think it would be possible to input their first and last name, as well as their github id? Sometimes it's a little difficult to track down who the github id belongs to. This could be a relatively easy solution? @hdamker @wrathwolf thoughts on this idea?

@eharrison24 That is of scope for this PR, but more or request or even potential policy towards members of the CAMARA Organization to update their GitHub Profile with appropriate, professional information. For the CODEOWNERS file the Github ID is sufficient, especially if it is already a member of the organization, e.g. in this case as Maintainer of the same subproject. I wouldn't duplicate any information in this file.

eharrison24:
@hdamker Makes sense! Appreciate the feedback.

eharrison24 (approved):

hdamker:
Added @maheshc01 to @camaraproject/edge-cloud_codeowners, so that the CODEOWNERS file gets valid.

Issue #314, opened 2024-10-30
TI API: user mobility problem


FabrizioMoggio:
**Problem description**
The TI API can be invoked for any user or for specific users. Currently if the API is invoked for any user and than it is invoked for a specific user, the "any user intent" is automatically deleted and the network is optimized only for the specifically requested users. This doesn't work with user mobility when I invoke the API for any user but then I detect that a user is moving among geographical areas. To optimize the network for that user I invoke the TI API for that user specifically and the system automatically delete the "any user intent". This behavior is not correct.

**Expected behavior**
The "any user intent" must be manually deleted by the API Consumer.

Use Case 1: optimization for any user while supporting user mobility
- TI API invoked for any user. Any user attached to an Edge UPF is routed to the Edge Application
- TI API invoked for the user that is moving. the user is moved to the closest UPF in that area
- The network is anyway configured to support any user

Use Case 2: after a trial phase where every user is optimized, just some selected user are optimized
- TI API invoked for any user (trial phase)
- TI API invoked to delete the "any user" intent
- TI API invoked for the selected users
- The network is optimized only for the selected users

**Required Action**
Modification of the TI API documentation

FabrizioMoggio:
Fixed with: https://github.com/camaraproject/EdgeCloud/pull/315

FabrizioMoggio:
Merged

Issue #284, opened 2024-07-05
TI API: phone number from access token


FabrizioMoggio:
**Problem description**
according to the CAMARA Guidelines, the phone number must be retrieved by the API Producer from the access token. The Device parameter must be optional and if used a cross check with the access token must be done

**Expected behavior**
implement this behavior according to: https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#appendix-a-infodescription-template-for-device-identification-from-access-token

FabrizioMoggio:
Fixed with: https://github.com/camaraproject/EdgeCloud/pull/315

FabrizioMoggio:
Merged

Issue #283, opened 2024-07-04
TI API: Cloud Event notification


FabrizioMoggio:
**Problem description**
The TI API is asynchronous, for this reason uses callbacks to notify the API Consumer with the result of the actions. Commonalities defines Cloud Event as the method for doing this. Is the TI API compatible with Cloud Event?

**Expected behavior**
The TI API must use Cloud Event to notify the API Consumer

**Additional context**
First of all we need to verify if we are compatible or not

FabrizioMoggio:
Reference to documentation on Cloud Event: https://github.com/camaraproject/Commonalities/blob/main/artifacts/notification-as-cloud-event.yaml

FabrizioMoggio:
Fixed with: https://github.com/camaraproject/EdgeCloud/pull/315

FabrizioMoggio:
Merged

Issue #246, opened 2024-05-23
TI API: 3Legs problem for optional parameter


FabrizioMoggio:
**Problem description**
The TI API adopted 3Legs authentication because it can provide back some information about the local of a device as a feedback of the traffic being influences in a certain zone or region. Anyway the device information is an option, the TI API can also be invoked in a generic way effecting all the connection toward an application

**Expected behavior**
3legs is expected when a device info is passed and it is not expected when a device is not provided as input. Currently Device is an optional parameter for the /traffic-influences POST method. The "security schema" for that method is "openId" while actually if the Device parameter is not used it should just be client credential. Is there a way to manage this situation with just one operation?

**Alternative solution**
this is solved if we create two operations:
1) POST with openId to create a TI Resource for a Device
2) another POST with client/credential to create a generic TI Resource for any Device

FabrizioMoggio:
Also the Patch method has the same problem. When no device is used no Consent management can be performed. So also two PATCH methods must be used, one with Client/Credential and one with OpenId

FabrizioMoggio:
Asked to Commonalities: https://github.com/camaraproject/Commonalities/discussions/245

FabrizioMoggio:
Asked in ICM: https://github.com/camaraproject/IdentityAndConsentManagement/discussions/184

**Conclusion**: two operations:

- POST with openId to create a TI Resource for a Device
- another POST with client/credential to create a generic TI Resource for any Device

FabrizioMoggio:
Merged

PR #315, opened 2024-11-04
TI API: Fall24 Alignment

FabrizioMoggio:
This PR is meant to align the TI API to the Fall24 guidelines addressing the following problems: 

- The TI API adopts 3Legs for any endpoints while sensitive data are not always exchanged.
- The description of the usage of the API for "user mobility" related scenarios is not good enough.
- The TI API doesn't not consider the phone number retrieved from the access token
- The TI API is not compliant with the CAMARA Guidelines in terms of Error Codes 
- The callback is not compliant with Cloud Event

#### What type of PR is this?

* bug
* correction
* enhancement/feature
* cleanup
* documentation

- This PR modifies the TI API adopting 3Legs only when sensitive data are exchanged. 
- This PR improve the API documentation better explaining how to use the API in "user mobility" scenarios
- This PR consider the phone number retrieved from the access token
- This PR adds the missing Error Codes 
- This PR implements the callback as Cloud Event

- User mobility problem: https://github.com/camaraproject/EdgeCloud/issues/314
- 3Legs problem for optional parameter: https://github.com/camaraproject/EdgeCloud/issues/246
- Phone number from access token: https://github.com/camaraproject/EdgeCloud/issues/284
- Error codes update: https://github.com/camaraproject/EdgeCloud/issues/320
- Cloud Event notification: https://github.com/camaraproject/EdgeCloud/issues/283


FabrizioMoggio:
this PR aligns the TI API to Fall24 (even if not included). Before moving to work on an alignment to Spring25 (again, even if not included in that Meta Release). It is better to close and merge this PR. Is there any objection or comment on this PR? If not, can it be approved?

Issue #255, opened 2024-06-07
EAM: Suggest to add App Instance Name


gainsley:
**Problem description**
No human-readable identifier for application instances. Very difficult differentiate between application instances only via UUID.

**Possible evolution**
Suggest to add a user-assigned `name` string field to application instances, much like AppManifest has a user-defined name field. This will allow users to easily distinguish between the different application instances they have deployed.

**Alternative solution**
None.

**Additional context**
None.

javierlozallu:
The UUIDs of the application instances are generated by the Edge Platform depending on the Edge Cloud Zones, so it is transparent for the Application Provider and this new parameter should follow the same approach, if we want to add this new parameter we should agree on a data model name for it as well (e.g. AppName-EdgeCloudZoneName).

Some doubts here:
- For one application can you deploy one app instance per EdgeCloudZone or several app instances per EdgeCloudZone? if multiple instances are allowed then we should add numbering to the data model name. 
- The multiple app instances option is based on auto scaling or is something that the app provider can define?
- An application could be composed of multiple microservices, how can we define a naming for this?

gainsley:
My intent was the name field be user-specified, not server generated. This is the same as AppManifest.Name. The user can choose any name they want, because that name will help them identify the application instance, so there is no specific format beyond being a string.

gainsley:
Regarding your doubts:
- We allow multiple instances of the same app per edge-cloud-site, but it's really irrelevant here. If the user chooses a name that already exists (and here we need to define the scope, i.e. the name must be unique per edge-cloud-site and per AppProvider, but not necessarily unique globally), then the API fails and the user must choose a non-conflicting name.
- Typically autoscaling is left to Kubernetes, and so all pods that are scaled are still part of a single app instance from the platform's point of view. In any case the name is for the app instance, not individual pods or VMs that may be scaled as part of the single app instance.
- Similarly, the name is for the app instance, not the individual microservices that may be part of the instance, so it doesn't matter how many individual services an app instance may be composed of.

javierlozallu:
> My intent was the name field be user-specified, not server generated. This is the same as AppManifest.Name. The user can choose any name they want, because that name will help them identify the application instance, so there is no specific format beyond being a string.

[...]

PR #316, opened 2024-11-06
EAM API: singular zone declarative app instance post API

gainsley:
#### What type of PR is this?

* correction
* enhancement/feature

- Deploy an AppInstance to a single Zone instead of multiple zones. This avoids confusion about the behavior if one or more of the multiple zone deployments fails, see https://github.com/camaraproject/EdgeCloud/issues/256.
- Assign a name to the AppInstance. Without this we cannot have declarative/infrastructure-as-code clients layered on top of the NBI API. See https://github.com/camaraproject/EdgeCloud/issues/255. Also makes it consistent with other objects that also have names, i.e. Apps, EdgeCloudZones.
- Adds more necessary information to the returned AppInstanceInfo, including the AppId (previously no way to figure out which App the AppInstance was an instance of). Makes the necessary information required. Changed the EdgeCloudZoneName to an EdgeCloudZoneId for consistency, as we always reference other objects by their ID, not their name.

Fixes https://github.com/camaraproject/EdgeCloud/issues/256, https://github.com/camaraproject/EdgeCloud/issues/255

#### Changelog input

```
 release-note
- appinstance post API targets a single zone instead of multiple zones and requires a name for declarative and infrastructure-as-code clients, added more information to appinstanceinfo

```

#### Additional documentation 

```
docs

```


LuisVTazon:
The original definition of the API is designed to allow the consumer of the API This proposal is intended to allow the user of the /appinstances resource to request the deployment of the same applicationin different zones, with one single call. 

This proposal seem to modify that behaviour by only allowing the deployment of an application in one zone per single request.

I think the original behaviour should stull be possible to simplify the user of the API, therefore I identify the following  potential options to this 

1. Create a new resource, different than /appinstances to manage a single deployment per request
2. Leaving the original  definition of /appinstances but modify the response so that it provides the status of each of the deployments that are triggered upon the request (zone and status of execution of request)

JoseMConde:
I have the same concern than @LuisVTazon commented, the idea is to make things easy to the use of the API, with this change we are obligating to the user to deploy these applications one by one per each zone.

gainsley:
I'm not sure it would necessarily be easier. For example, if you are deploying through a web UI, the user will probably check a bunch of check boxes for which zones to deploy to, and then hit a deploy button. Whether the UI code runs one API call or many, the user doesn't actually know. In fact, individual calls may be better, in case some finish earlier than others, the user gets feedback sooner, rather than having to wait until all AppInstance requests are processed.

How are you imagining these APIs will be used?

Additionally, the DELETE API is for a single instance, so anyway DELETE must be managed one-by-one. It seems inconsistent that the API allows me to create many at once, but only delete one at a time. Why is this optimization necessary for create, but not for delete?

If we have a multi-zone API call:
- what is the overall response status code if some AppInsts are accepted (202), some fail (400), and some have conflicts (409) or other statuses?
- if a target zone already has an instance of the Application, is the intent that we create another instance, or is this a conflict (409), or do we accept and return the ID of the already existing AppInst?

JoseMConde:
Hi @gainsley ,
As far as I know, the idea of Camara APIs is not be exposed by web UI, the idea is that APIs should be call directly by the developer without web UI.
In my view, is different when we talk about delete or deploy, I mean, if you by mistake deploy several Instances instead of just one, this doens´t impact in the final user, but if you delete several Instances by mistake, this could impact in the final user, for this reason, we didin't allow the developer to delete several Instances in one call.
About your questions, you are totally wright, we need to improve this.
For the first question, I think the response could be an array of responses, something like this:
One response per AppInsts.
- 202, Application instantation accepted.
- 202, Application instantation accepted.
- 202, Application instantation accepted.
- 202, Application instantation accepted.
- 400, bad request.
- 202, Application instantation accepted.
For the second question, I think It should return 409, just to let the developer know that this instances is already Instantiated.

gainsley:
Hi @JoseMConde, so my first question still needs a decision, the API must have a single HTTP response code as part of the HTTP protocol. We can of course embed response codes per AppInst inside the body of the response, but that's orthogonal to the HTTP protocol as it's just body data. The HTTP protocol still requires a single response code - so is that code always 200 or 202, or is it based on what the per-AppInst response codes are? I think we'll basically need to tell the user to ignore the HTTP protocol response code, and instead look at the response codes in the response JSON body.

Additionally, please consider that each AppInst request needs a name, potentially a cluster ID, and in the future may also want to include more parameters, like a helm values file. So the request will be an array of AppInst requests, i.e. an array where each item is:

```
- name
- clusterID
- zoneID
- instance specific parameters (helm values, etc)
```

If I were running curl requests by hand, I would imagine it's easier to craft a single AppInst request, then copy-paste-edit it for each AppInst, rather having to a craft a single big JSON array by hand to send everything in one request.

JoseMConde:
Hi @gainsley , Thanks for the comment.
In my view, the response code should be always 202, and as you perfectly commented this has to be the approach: 

> I think we'll basically need to tell the user to ignore the HTTP protocol response code, and instead look at the response codes in the response JSON body.

>
About the second point, you can also copy and paste inside the json array to make the same request, I mean, it would be almost the same working time, or am I missing anything??

gainsley:
@JoseMConde I have created an alternative PR #333 which adheres to the multi-request approach, with the same necessary fixes from this PR. I suggest we review that PR and have people decide which approach they prefer between #316 and #333.

JoseMConde:
Thanks @gainsley, totally agree. I think is the best way.

JoseMConde:
@camaraproject/edge-cloud_codeowners Please, take a look to this PR and give some feedback.

gunjald:
I also think supporting application deployment to multiple edge cloud zone in one request should be possible as it can also serve the scenario where a single instance is to be created. The status code 202 though should be augmented with additional GET/DELETE methods to later query or remove individual application request objects.

JoseMConde:
Thanks @gainsley @gunjald Do you think could make sense to close this PR and still discussing in [PR-333](https://github.com/camaraproject/EdgeCloud/pull/333) or do you still have doubts about which is the optimal way to proceed??

PR #318, opened 2024-11-06
EAM API: add get clusters api

gainsley:
#### What type of PR is this?

* enhancement/feature

Addresses a comment in #280 regarding the addition of the KubernetesClusterRef, and how users would expect to find it. While the KubernetesClusterRef can be found in the AppInstanceInfo, there may be clusters present without any AppInstances. The only way to see those clusters is via a cluster API to list the clusters.

#### Changelog input

```
 release-note
- add clusters list API

```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -547,6 +547,68 @@ paths:
           $ref: '#/components/responses/500'
         '503':
           $ref: '#/components/responses/503'
+  /clusters:
+    get:
+      security:
+        - openId:
+            - edge-application-management:clusters:read
+      tags:
+        - Cluster
+      summary: |
+        Retrieve a list of the available clusters filtered by the optional
+        query parameters.
+      description: |
+        List available cluster information
+      operationId: getClusters
+      parameters:
+        - $ref: '#/components/parameters/x-correlator'
+        - name: region
+          description: |
+            Human readable name of the geographical Edge Cloud Region of
+            the Cluster. Defined by the Edge Cloud Provider.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/EdgeCloudRegion'
+        - name: clusterRef
+          description: |
+            A globally unique identifier for the Cluster.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/KubernetesClusterRef'
+        - name: edgeCloudZoneId
+          description: |
+            Edge Cloud Zone identifier.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/EdgeCloudZoneId'
+      responses:
+        '200':
+          description: |
+            Successful response, returning the cluster's information.
+            Returns an empty list if no clusters were found or none match
+            the specified query parameters.
+          headers:
+            x-correlator:
+              $ref: "#/components/headers/x-correlator"
+          content:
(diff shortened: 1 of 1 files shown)


code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -547,6 +547,59 @@ paths:
...
+            - edge-application-management:clusters:read
+      tags:
+        - Cluster
+      summary: Retrieve a list of the available clusters
+      description: |
+        List available cluster information
+      operationId: getClusters
+      parameters:
+        - $ref: '#/components/parameters/x-correlator'
+        - name: region
```
gunjald, line 563: As the EdgeCloudZone has been used as the prime factor to refer to edge resources and region as the secondary attribute so should we include the EdgeCloudZone and optionally the region?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -547,6 +547,59 @@ paths:
...
+          description: |
+            A globally unique identifier for the Cluster.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/KubernetesClusterRef'
+      responses:
+        '200':
+          description: |
+            Successful response, returning the clusters.
```
gunjald, line 581: I suppose returning the clusters information?

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -547,6 +547,59 @@ paths:
...
+            the Cluster. Defined by the Edge Cloud Provider.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/EdgeCloudRegion'
+        - name: clusterRef
+          description: |
+            A globally unique identifier for the Cluster.
+          in: query
+          required: false
```
gunjald, line 575: Is it correct that if the clusterRef is not present in query than the response will include information of all the clusters in that edge cloud zone and/or region. If it is correct then we can include this indication in description. To me we need this capability as there is no other method by which I as a developer can retrieve the clusterRef.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -907,6 +960,25 @@ components:
       pattern: ^[A-Za-z][A-Za-z0-9_]{7,63}$
       description: Human readable name of the Application Provider.
 
+    ClusterInfo:
```
gunjald, line 963: Although the ClusterInfo object returns the additional metadata about the clusterRef but shouldn't there be a bit more about the cluster itself like number of nodes, k8s version, GPU enabled etc?

code/API_definitions/Edge-Application-Management.yaml
gainsley: Yes, that's the intent, as stated in the summary. I'd argue it's somewhat obvious since these query parameters are optional, but I can make it more clear in the summary.

code/API_definitions/Edge-Application-Management.yaml
gainsley: @gunjald I have updated this, please take a look. thanks.

gunjald: Looks fine to me.

gainsley:
All reviewer comments have been addressed, I believe this can be merged.

JoseMConde (approved):

PR #319, opened 2024-11-06
EAM API: clarify appinstances get behavior

gainsley:
#### What type of PR is this?

* correction

Clarifies the behavior of the GET appinstances API if there are no matching instances to return. Implementors may consider returning 404 in this case, but I believe it is more correct to return 200 with an empty list, because:
- 404 is also returned if the URL path is incorrect, which makes it hard to distinguish between no matches vs invalid URL path
- Users (in particular tests) may intentionally be checking for the absence of an instance. Returning 404 in this case complicates verifying the absence of an instance.

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -479,7 +479,10 @@ paths:
             $ref: '#/components/schemas/EdgeCloudRegion'
       responses:
         '200':
-          description: Information of Application Instances
+          description: |
+            List of application instances. Returns an empty list if no
+            instances were found or none match the specified query
+            parameters.
           headers:
             x-correlator:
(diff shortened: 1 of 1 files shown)


albertoec (commented):
Hi @gainsley ,

I completely agree that for an empty list, the correct response should be a 200 HTTP status code with an empty array, rather than a 404. However, there are several other models (e.g., the response for GET /edge-cloud-zones) that currently behave this way, and this behavior isn’t explicitly specified.

I think we should decide whether we want to consistently specify the behavior for empty arrays in all responses.

What do you think?

gainsley:
@albertoec I agree it should be consistent. I would suggest merging this change and a follow-up PR can add the same description to other APIs.

JoseMConde (approved):

Issue #286, opened 2024-07-22
EDS API: Application Endpoints functionality to be a new API


javierlozallu:
**Problem description**
This Operation allows to: Register and manage the routable Application Endpoints of your deployed applications.
Currently the approach is that endpoints are fully managed by the Edge Platform and Developers can ask for this information through GET /apps/{appId}/instances EAM API method.
I don't found any intent related to this capability but I think is very interesting to have it because of two reasons:

-     If the developer deploys the application on a hyperscaler
-     If the developer wants to use their own domains, this relates to a comment in 

Shouldn't the AccessEndpoints also contain the protocol  #271 to have a new API: Bring Your Own DNS/Endpoint/Whatever which allows the developer to use and manage third party domains (other than the Edge platform). So this contribution will serve as a basis for this.
A different approach I see here is to add it in the onboarding process within the EAM API, but I don't quite like it.

**Expected behavior**
Create separate yaml for Application end point registry. This is more edge cloud specific but still reusable across the sub projects.

maheshc01:
PR created for this with initial set of API capabilities. More updates regarding API documentation and standardizing of error codes to be updated as part of the same PR.

Kevsy:
@maheshc01 was the PR mentioned above merged and hence the issue can be resolved?

maheshc01:
yes. the PR was merged. will go ahead and close this issue.

PR #321, opened 2024-11-11
application end point registration API yaml

maheshc01:
#### What type of PR is this?
This PR is to review and update application endpoint registration API yaml. 

* bug
* correction
* enhancement/feature
* cleanup
* documentation
* subproject management
* tests

Initial commit of the yaml to provide ability to register application endpoints.

issue #286 

Fixes #

#### Special notes for reviewers:
This is only the initial commit and additional updates are required related to documentation, error codes etc to ensure it meets the CAMARA guidelines.

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```


code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
+openapi: 3.0.3
+info:
+  title: Application Endpoint Registration API
+  version: 0.1.0
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  description: |
+    Application endpoint registration allows application developers to
+    register one or more Application Endpoints, and retrieve, update,and
```
gunjald, line 10: Should we have a definition for "Application Endpoints" to make it more explicit?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+    delete those egistrations.
```
gunjald, line 11: I suppose egistrations is should be "registrations"

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+
+    This information can we used for various usecases like optimal end
+    point discovery to help end users connect to the most most optimal
+    instance of the application. Addtionally the information can be used
```
gunjald, line 15: I feel that if can refer or qualify applications as "edge applications" then it will make it more related to edge deployments while using generically may imply any deployment including clouds. Just a thought.

code/API_definitions/Application-endpoint-registration.yaml
gunjald: I also think that these Application Endpoints are not generic and they corresponds to the application instances running on edge cloud zones in probably distributed manner. So should we qualify the Application Endpoints in some way to indicate developers about the expectations from these endpoints i.e. edge instances?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  description: |
+    Application endpoint registration allows application developers to
+    register one or more Application Endpoints, and retrieve, update,and
+    delete those egistrations.
+
+    This information can we used for various usecases like optimal end
+    point discovery to help end users connect to the most most optimal
```
gunjald, line 14: "most most optimal" should be changed to "most optimal"

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+        description: Array of Application Endpoints for a
+          deployed application
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: "#/components/schemas/ApplicationInstance"
+      responses:
+        "200":
```
gunjald, line 71: We should also consider the additional status codes which could be valid for the POST requests in alignment with other APIs and consistency.

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+tags:
+  - name: Application Endpoint Registration
+    description: |
+      Operations to register, read and manage an deployed instances of the
+      application.
+
+paths:
+  /application-endpoints:
+    post:
+      operationId: register-application-endpoints
```
gunjald, line 60: For the POST method should we also include "x-correlator" header as done with other APIs.?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+        edgeCloudZoneId:
+          $ref: "#/components/schemas/EdgeCloudZoneId"
+        edgeCloudZoneName:
+          $ref: "#/components/schemas/EdgeCloudZoneName"
+        edgeCloudProvider:
+          $ref: "#/components/schemas/EdgeCloudProvider"
+        edgeCloudRegionId:
+          $ref: "#/components/schemas/EdgeCloudRegionName"
+    ResourcesapplicationEndpoint:
+      type: object
```
gunjald, line 205: Should we assume that an application can expose only one endpoint or there could be more than one for different purposes? If there can be more than one which only application provider knows then this parameter should allow the multiplicity by having it as an array of objects. Also in that case with each endpoint there should be a discriminator field to allow client applications to know the purpose of a specific endpoint.

gunjald: Although the top level object is an array but then if application has multiple endpoints some of the information will be repeated that many times as they will be common to an application.

gainsley: I also had the same question about repetitive data in the array. For example, is it expected that applicationId is the same for each ApplicationInstance in the array? If any field must be the same for each item in the array, it should be factored out into a single field in a parent object that encapsulates the endpoints array. If every field in the array can be different, then what is the intent of assigning a single applicationEndpointsId to the entire array?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+      summary: Register Application Endpoints
+      description: Register Application Endpoints of a
+        deployed application to specified edge cloud zone.
+
+  "/application-endpoints/{applicationEndpointsId}":
+    parameters:
+      - name: applicationEndpointsId
+        in: path
+        required: true
+        description: applicationEndpointsId param //added desc
```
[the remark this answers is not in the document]
urvika-v: Fixed

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+          `api.example.com` or `api.example.com/somepath`
+
+tags:
+  - name: Application Endpoint Registration
+    description: |
+      Operations to register, read and manage an deployed instances of the
+      application.
+
+paths:
+  /application-endpoints:
```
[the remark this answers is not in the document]
urvika-v: Agreed. Added "Get" which lists all the registered applications.

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+      apiRoot:
+        default: http://localhost:9091
+        description: |
+          API root, defined by service provider, e.g.
+          `api.example.com` or `api.example.com/somepath`
+
+tags:
+  - name: Application Endpoint Registration
+    description: |
+      Operations to register, read and manage an deployed instances of the
```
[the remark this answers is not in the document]
urvika-v: Modified to "read and manage deployed instances"

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+      application.
+
+paths:
+  /application-endpoints:
+    post:
+      operationId: register-application-endpoints
```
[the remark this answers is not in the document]
urvika-v: Fixed.

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+      type: string
+    TypesApplicationEndpointsId:
+      description: |
+        A system-defined string identifier representing
+        one or more registered Application Endpoints.
+      type: string
+      format: uuid
+      readOnly: true
+      additionalProperties: false
+    TypesApplicationProfileId:
```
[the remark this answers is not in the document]
urvika-v: Modified "TypesApplicationProfileId" to "ApplicationProfileId"

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+      additionalProperties: false
+    EdgeCloudProvider:
+      description: |
+        The company name of the Edge Cloud Zone provider.
+      type: string
+    EdgeCloudRegionName:
+      description: |
+        region of the Edge Cloud Zone.
+      type: string
+    TypesApplicationEndpointsId:
```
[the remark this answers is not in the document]
urvika-v: Fixed

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+    x-correlator:
+      description: Correlation id for the different services
+      schema:
+        type: string
+
+  schemas:
+    ApplicationInstance:
+      description: Application instance represented
+        by application Endpoint definition
+      type: object
```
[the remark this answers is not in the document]
urvika-v: Added "Required" section to the schema.

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Included  "x-correlator" for all the APIs

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "404":
+          $ref: "#/components/responses/NotFound"
+        default:
+          $ref: "#/components/responses/Unexpected"
+      tags:
+        - Application Endpoint Registration
+      summary: Update a application Endpoint
+      description: Update registered application Endpoint information.
```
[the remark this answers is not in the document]
urvika-v: Moved the summary and the description to the top.

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: > Should we have a definition for "Application Endpoints" to make it more explicit?

Added the "Application Endpoints" definition in "Relevant terms and Definitions" section

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: modified

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Yes, modified "applications" to "edge applications"

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Modified

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Added relevant status codes for all the APIs

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+        applicationProfileId:
+          $ref: "#/components/schemas/TypesApplicationProfileId"
+        edgeCloudZoneId:
+          $ref: "#/components/schemas/EdgeCloudZoneId"
+        edgeCloudZoneName:
+          $ref: "#/components/schemas/EdgeCloudZoneName"
+        edgeCloudProvider:
+          $ref: "#/components/schemas/EdgeCloudProvider"
+        edgeCloudRegionId:
+          $ref: "#/components/schemas/EdgeCloudRegionName"
```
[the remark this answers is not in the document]
urvika-v: Yes, the edgeCloudZone-related fields are to be specified during the POST call. And it is expected that the Operator platform should verify whether its a valid cloud zone or not before the "ApplicationEndpointsId" gets created

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: In this initial version, the API handles where an application has a single purpose but distributed across various edge locations so each application instance at a given edge location has an endpoint to register. 
The scenario mentioned application endpoints with different purpose is also a valid one. This can be considered in the  later release.

gunjald: The title "title: Application Endpoints Registration" has plural "Endpoints" as the keyword and based on the current form with single endpoint we may change it to singular form as endpoint or if we want to handle endpoints with different purposes in this release only then we need to provide option for multiple endpoint i.e. array in my view.

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,625 @@
...
+    Additionally the information can be used to monitor the edge instances to
+    take decisions from a lifecycle management perspective. The API provides
+    the ability to register, read and manage the deployed edge instances of
+    the application.
+
+    # Relevant terms and definitions
+
+    * **Application Endpoint**:
+
+    The Endpoint on which the application is running.
```
gunjald, line 26: Can we slightly update the statement "application is running" to such as "application is accessible"?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,625 @@
...
+    It can be a URI, FQDN, IPv4, or IPv6 address with a port number.
```
gunjald, line 27: Should the protocol i.e. TCP/UDP also be part of the endpoint spec?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,625 @@
...
+        deployed across various cloud zones.
+      properties:
+        applicationEndpointsId:
+          $ref: "#/components/schemas/ApplicationEndpointsId"
+        applicationInstance:
+          description: |
+            Application Endpoint Information
+          type: array
+          items:
+            $ref: "#/components/schemas/ApplicationEndpointRequest"
```
gunjald, line 383: The term ApplicationEndpointRequest does not aligned to the top-level object name ApplicationEndpoins. May it could be ApplicationEndpointInfo or something like that?

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,625 @@
...
+        - openId:
+            - application-endpoint-registration:application-endpoints:read
+      tags:
+        - Application Endpoint Registration
+      summary: Get all registered edge application endpoint information
+      description: |
+        Returns endpoint information for all
+        Application Endpoints registered.
+      operationId: getAllRegisteredApplicationEndpoints
+      parameters:
```
gunjald, line 164: I was thinking from the usage perspective of GET method on how the apps suppose to use it. To me it it look like there will be a case where applications would need to retrieve the endpoint information of given application in a given location. As of, in current structure it seems the application will need to iterate through the whole array and find out which is the right endpoint of an app.
So if this seems to be a valid use case then should we have query parameters for say appId and/or edge cloud zone as an example? Though with the two GET method the same goal may still be achieved but i was just thinking if we should provide additional options for such endpoint retrieval scenarios?

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Fixed.

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: The intent of this API is registration and management of the Application endpoints. There is a support to retrieve the Application Endpoints based on the AppID. But  retrieval of the right/optimal Application endpoint is part of Discovery API. This would require more discussion and take a call based on the Discovery API filter options

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Modified "ApplicationEndpointRequest" to "ApplicationEndpointInfo". Also just a note I am reusing this schema for other APIs.

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: The Application would know what protocol its using. In this case it would not require as the goal is to identify the Optimal Application Endpoint.

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: "Application Endpoints" and its type "array" is still valid in this initial version because the same application can be deployed across multiple edge locations.

gunjald: So does this mean ResourcesapplicationEndpoint should be converted to an array in some way? Is that the right interpretation of your comment?

gunjald: I think this object is part of ApplicationEndpoints in the form of array. In that case my previous comment will become invalid.

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,625 @@
...
+    # Relevant terms and definitions
+
+    * **Application Endpoint**:
+
+    The Endpoint on which the application is running.
+    It can be a URI, FQDN, IPv4, or IPv6 address with a port number.
+
+    # API functionality
+
+    The API provides 4 operations:
```
JoseMConde, line 31: 5 Operations now?? :)

code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -0,0 +1,392 @@
...
+          $ref: "#/components/schemas/TypesApplicationEndpointsId"
+        applicationEndpoint:
+          $ref: "#/components/schemas/ResourcesapplicationEndpoint"
+        applicationServerProviderName:
+          type: string
+          description: Unique ID representing the Edge
+            Application Provider
+        applicationId:
+          type: string
+          description: Unique ID representing the Edge Application
```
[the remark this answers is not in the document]
maheshc01: I believe this point needs more discussion. we need provide developers will more clarity and details on how these different APIs in edge cloud work togethers. EAM used for orchestrating the individual instances of the application. Application end point registration is a mechanism for creating a database of records for all the active and available application instances and the same information will be using in traffic influence and application end point discovery to identify the specific application instance. I think we need to take a closer look at how and where we are using the terms appId and appInstanceId

JoseMConde:
Hi @urvika-v , there are still some comments to anwser, at least my last one and I've seen others for Deepak and Jon, could you please review it??

code/API_definitions/Application-endpoint-registration.yaml
urvika-v: Yes, it will be 5 operations.

JoseMConde: So, we should update the YAML according that, shouldn't we??

urvika-v: Modified the yaml and created PR for it.

urvika-v:
The linting errors raised in this PR are related to the Application_Endpoint_Discovery_API.yaml file which got missed in the PR: https://github.com/camaraproject/EdgeCloud/pull/337. As suggested , we are gonna close this PR and create a new PR for Application Endpoint Registration

PR #322, opened 2024-11-14
EAM API: make GET appinstances response data consistent

gainsley:
#### What type of PR is this?

* correction

This removes an unnecessary layer of hierarchy in the GET appinstances response, so it returns an array of objects rather than an array of objects encased in a parent object. This also makes it consistent with GET apps and GET edge-cloud-zones which also return arrays of objects, not arrays of objects inside an object.

#### Changelog input

```
 release-note
- remove enclosing object in GET appinstances response
```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -486,13 +486,9 @@ paths:
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


gunjald (approved):
To me the changes looks good.

JoseMConde (approved):

PR #326, opened 2024-11-25
Edge discovery APIs enhancements

kohlian18:
#### What type of PR is this?
* enhancement/feature
* cleanup
* Documentation

This segregates and simplifies edge cloud zone and application endpoint discovery APIs:

- GET edge cloud regions API to fetch edge cloud region id and name required for discovery APIs.
- POST method containing standardised device information schema along with application profile id for edge cloud zone discovery.
- POST method along with device information with application endpoint ids and edge cloud regions for application endpoint discovery.
- added information for access and authorisation token

### DIFF
--- a/code/API_definitions/edge-discovery-service.yaml
+++ b/code/API_definitions/edge-discovery-service.yaml
@@ -250,9 +250,9 @@ tags:
 #                                     Paths                                #
 ############################################################################
 paths:
-  /regions:
+  /edge-cloud-regions:
     get:
-      summary: Get all Region IDs and Names
+      summary: Get all Edge Cloud Region IDs and Names
       operationId: get-region-ids-and-names
       parameters:
         - name: x-correlator
@@ -388,7 +388,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Device"
+                $ref: "#/components/schemas/EdgeApplicationEndpointsDevice"
         responses:
           '200':
             description: Find optimal Application Endpoints for clients to connect to
@@ -444,7 +444,12 @@ paths:
         tags:
         - Discovery
         summary: Find optimal Application Endpoints for clients to connect to
-        description: Returns a list of optimal Application Endpoints that client devices can connect to. You can search based on Application Profile and device information
+        description: |
+          Returns a list of optimal Application Endpoints that client devices can connect to. 
+          You can search using a single application endpoint id or a list of application 
+          endpoint ids or combination of application endpoint ids along with device information.
+          Edge cloud region name can also be used as another filter to find the optimal 
+          application endpoints.
 
 components:
   ######################################################
@@ -671,12 +676,44 @@ components:
         
         applicationProfileId:
           $ref: "#/components/schemas/ApplicationProfileId"
-        edgeCloudRegion:
-          $ref: "#/components/schemas/EdgeCloudRegionName"
       required:
         - applicationProfileId
         
+    EdgeApplicationEndpointsDevice:
+      description: |
+        End-user equipment able to connect to a mobile network. Examples of devices include smartphones or IoT sensors/actuators.
 
+        The developer can choose to provide the below specified device identifiers:
+
+        * `ipv4Address`
+        * `ipv6Address`
+        * `phoneNumber`
+        * `networkAccessIdentifier`
+
+        NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
+      type: object
+      properties:
+        device:
+          type: object
+          properties:
+            phoneNumber:
+             $ref: "#/components/schemas/PhoneNumber"
+            networkAccessIdentifier:
+             $ref: "#/components/schemas/NetworkAccessIdentifier"
+            ipv4Address:
+             $ref: "#/components/schemas/DeviceIpv4Addr"
+            ipv6Address:
+             $ref: "#/components/schemas/DeviceIpv6Address"
+        
+        applicationEndpointsIds:
+          type: array
+          items:
+            $ref: "#/components/schemas/ApplicationEndpointsId"
+        edgeCloudRegionName:
+          $ref: "#/components/schemas/EdgeCloudRegionName"
+
+      required:
+        - applicationEndpointsIds
 
     PhoneNumber:
       description: A public identifier addressing a telephone subscription. In mobile networks it corresponds to the MSISDN (Mobile Station International Subscriber Directory Number). In order to be globally unique it has to be formatted in international format, according to E.164 standard, prefixed with '+'.
--- a/code/API_definitions/edge-discovery-service.yaml
+++ b/code/API_definitions/edge-discovery-service.yaml
@@ -11,14 +11,198 @@ info:
   description: |
     # Introduction
     ---
-    Use the Edge Discovery API for
-    discovery of the most optimal edge cloud zone for a specific user as well as specific application instance the user is trying to connect to.
+    The Edge Discovery API is designed to help users discover the most optimal
+    edge cloud zones for their applications. This API allows users to find the
+    best regions and zones for their edge cloud applications, discover optimal
+    edge cloud zones for deployed applications, and find the best application
+    endpoints for clients to connect to.
 
-    ---
+    ## Features
+    - Discover regions and zones to deploy edge cloud applications.
+    - Discover optimal application endpoints for client connections for deployed applications.
+
+    ## Usage
+    Users can utilize this API to enhance the performance and connectivity of
+    their applications by selecting the most suitable edge cloud zones and
+    application endpoints.
+
+    ## Use Cases
+    - **Edge Cloud Zone Discovery**: A developer wants to find the best edge cloud zones to deploy their application to minimize latency for users in a specific region.
+    - **Application Endpoint Discovery**: A client application needs to find the optimal endpoints to connect to for the best performance and reliability.
+    
     # API Scope
+    
+    The APIs defined in this version of the specification are categorized into
+    the following areas:
+    - Edge Cloud Zone Discovery
+    - Application Endpoint Discovery
+    
+    # Identifying the device
+
+    The API returns the closest Edge Cloud Zone to a given device, so that
+    device needs to be identifiable by the network. This can be achieved either
+    by passing a device identifier in the request header, or, from the 3-legged
+    access token where implemented by the operator.
+
+    ## Passing a device identifier in the request header
+
+    * If you call the API from a server, you must explicitly pass one or
+    more device identifiers in the HTTP request header:
+      * `IPv4-Address` or `IPv6-Address`. This is the public IP address of the
+      user device: this
+      can be obtained by an application hosted on that device calling a
+      public IP address API (e.g. GET https://api.ipify.org?format=json)
+      * If you provide an `IPv4-Address` or `IPv6-Address`: for certain
+      operators you may be required to also provide a `Public-port` header.
+      * `Phone-Number` . The international E.164 format (starting with country
+      code), e.g. +4407123123456
+      * `Network-Access-Identifier` (where available from the API host operator)
+
+    * If you call the API from a device attached to the operator network, then
+    you can attempt the request without passing device identifier(s) parameters
+    in the request header. If that returns a 422 `UNIDENTIFIABLE_DEVICE`
+    error then retry the request but this time include a device identifier.
+
+    NOTE1: the network operator might support only a subset of these options.
+    The API invoker can provide multiple identifiers to be compatible across
+    different network operators. In this case the identifiers MUST belong to the
+    same device.
+
+    NOTE2: for the Commonalities release v0.4, we are enforcing that the
+    `networkAccessIdentifier` is only part of the schema for future-proofing, and
+    CAMARA does not currently allow its use. After the CAMARA meta-release work
+    is concluded and the relevant issues are resolved, its use will need to be
+    explicitly documented in the guidelines.
+
+      ## Identifying a device from the access token
+
+    This specification defines the `device` identifying headers as optional in
+    API requests, specifically in cases where the API is accessed using a
+    3-legged access token and the device can be uniquely identified by the token.
+    This approach simplifies API usage for API consumers by relying on the device
+    information associated with the access token used to invoke the API.
+
+    ### Handling of device information:
+
+    #### Optional device identifying headers for 3-legged tokens:
+
+    - When using a 3-legged access token, the device associated with the access
+      token must be considered as the device for the API request. This means that
+      a device  identifying header  is not required in the request, and if
+      included it must identify the same device, therefore **it is recommended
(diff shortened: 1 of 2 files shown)


code/API_definitions/edge-discovery-service.yaml
```diff
@@ -11,14 +11,198 @@ info:
   description: |
     # Introduction
     ---
-    Use the Edge Discovery API for
-    discovery of the most optimal edge cloud zone for a specific user as well as specific application instance the user is trying to connect to.
+    The Edge Discovery API is designed to help users discover the most optimal
+    edge cloud zones for their applications. This API allows users to find the
+    best regions and zones for their edge cloud applications, discover optimal
```
gunjald, line 16: One generic recommendation would be to use "optimal" instead of "best" as we discussed this in past meetings as that may be hard to determine best edge for various reasons.

code/API_definitions/edge-discovery-service.yaml
```diff
@@ -11,14 +11,198 @@ info:
...
+    edge cloud zones for deployed applications, and find the best application
+    endpoints for clients to connect to.
```
gunjald, line 18: Need to refer to edge cloud zone instead of zone for consistency purpose as this is the formal term to indicate edge resources as I think.

code/API_definitions/edge-discovery-service.yaml
```diff
@@ -11,14 +11,198 @@ info:
   description: |
     # Introduction
     ---
-    Use the Edge Discovery API for
-    discovery of the most optimal edge cloud zone for a specific user as well as specific application instance the user is trying to connect to.
+    The Edge Discovery API is designed to help users discover the most optimal
+    edge cloud zones for their applications. This API allows users to find the
+    best regions and zones for their edge cloud applications, discover optimal
+    edge cloud zones for deployed applications, and find the best application
```
gunjald, line 17: I suppose this intent "discover optimal edge cloud zones for deployed applications" is valid for the end users which would connect with the edge applications instances. In that case we need to possibly indicate the different type of users who may consume these APIs.

code/API_definitions/edge-discovery-service.yaml
```diff
@@ -11,14 +11,198 @@ info:
...
+    endpoints for clients to connect to.
 
-    ---
+    ## Features
+    - Discover regions and zones to deploy edge cloud applications.
+    - Discover optimal application endpoints for client connections for deployed applications.
+
+    ## Usage
+    Users can utilize this API to enhance the performance and connectivity of
```
gunjald, line 25: It is bit confusing when it says the users can use this API to enhance connectivity and performance by selecting suitable edge cloud zone and application endpoints. In my view the app providers may need to use this API to deploy applications while the end users may use it just to discover application endpoints to connect with. 
So in terms of the usage it would be good to describe the scenario where the given intents are to be used by the intended users as per the goals of the API. Thoughts?

code/API_definitions/edge-discovery-service.yaml
kohlian18: resolved

code/API_definitions/edge-discovery-service.yaml
kohlian18: Resolved. Added two scenarios for api usage: Edge Cloud Zone Discovery and Application Endpoint Discovery

JoseMConde:
Should we update the name to Optimal Edge Discovery as we agreed in the last meeting??

kohlian18:
> Should we update the name to Optimal Edge Discovery as we agreed in the last meeting??

Sure, i will update it and push the changes.

JoseMConde:
Trying to see the API in a more generic view, I have some concerns:
1. The first point is that I still thinking that the methods should be Get not Post. The approach is almost the same than the Get/Edcloudzones that we have in EAM and also in AED, we can have a Get with some parameters inside the request body and some of them could be required and some of them not.
2. Another point is that `Post/Edge-cloud-zones`, doesn't have the same approach than `Get/Edgecloudzones` that we actually have in EAM. I think is interesting that the developer could filter by status. Also the `ApplicationProfileId` is a mandatory parameter, but maybe someone doesn't use Application Endpoint Registration and doesn't have this parameter or they use it for some Applications and for others they don't.
3. Analizing the behaviour of `POST/app-endpoint`s in this API and the behaviour of Application Endpoint Discovery, could be almost the same, the only thing taht is different is that we use `AppId` to know wich is the Application that we need the endpoint and this API uses `ApplicationEndpointsId`, am I wright or I am missing something??

maheshc01:
> Should we update the name to Optimal Edge Discovery as we agreed in the last meeting??

yes. the yaml needs to be renamed to Optimal Edge Discovery and remove the /application-endpoints from the yaml.

gainsley:
Hi @JoseMConde just a quick comment here: HTTP GET does not support a request body. You can pass parameters as URL parameters, but if you have a large complex set of a data (like JSON) that gets ugly pretty quickly.

maheshc01:
Please find my replies inline.

> Trying to see the API in a more generic view, I have some concerns:
> 
> 1. The first point is that I still thinking that the methods should be Get not Post. The approach is almost the same than the Get/Edcloudzones that we have in EAM and also in AED, we can have a Get with some parameters inside the request body and some of them could be required and some of them not.
[Response from @maheshc01 ]: using POST call and the device schema in the request body is the standard followed across all CAMARA APIs with the only exception of edge cloud based on what i know. Please check @hdamker's comments for simple edge discovery on the same in the PR for spring 2025 release. using POST and the device object is the CAMARA standard way. Please check Herberts comments in this PR(https://github.com/camaraproject/SimpleEdgeDiscovery/pull/66) specific to simple edge discovery service. we can also refer to other APIs like device status etc where the intents is to get information but POST and standard input body is use.

> 2. Another point is that `Post/Edge-cloud-zones`, doesn't have the same approach than `Get/Edgecloudzones` that we actually have in EAM. I think is interesting that the developer could filter by status. Also the `ApplicationProfileId` is a mandatory parameter, but maybe someone doesn't use Application Endpoint Registration and doesn't have this parameter or they use it for some Applications and for others they don't.
[Response from @maheshc01 ] i believe we discussed this during the call. based on what we agreed, Get/Edgecloudzones needs to be removed from AEM as it puts different intents related to orchestration and discovery into one yaml and as per our discussions we are separating it out to create yaml as per specific intents.
ApplicationProfileId is not from application end point registration. its will be from Application profiles API. the latest update on the application profile scope expansion is that its been discussed in api backlog group and is expect to be on the TSC agenda for approval soon. the parameter being mandatory, to discovery the optimal edge cloud zone we need additional information about the application and hence applicationProfileId is mandatory. If developers dont want to use this approach then they would use the simple edge discovery API.

> 3. Analizing the behaviour of `POST/app-endpoint`s in this API and the behaviour of Application Endpoint Discovery, could be almost the same, the only thing taht is different is that we use `AppId` to know wich is the Application that we need the endpoint and this API uses `ApplicationEndpointsId`, am I wright or I am missing something??
[Response from @maheshc01 ] as discussed during our meetings, this end point needs to be removed from this optimal edge discovery yaml and the related updated need to be done on AED yaml. we will need a different PR for this. @kohlian18  could you create a different PR for this.

JoseMConde:
Hi @maheshc01 ,
My thoughts:
> The first point is that I still thinking that the methods should be Get not Post. The approach is almost the same than the Get/Edcloudzones that we have in EAM and also in AED, we can have a Get with some parameters inside the request body and some of them could be required and some of them not.
[Response from @maheshc01 ]: using POST call and the device schema in the request body is the standard followed across all CAMARA APIs with the only exception of edge cloud based on what i know. Please check @hdamker's comments for simple edge discovery on the same in the PR for spring 2025 release. using POST and the device object is the CAMARA standard way. Please check Herberts comments in this PR(https://github.com/camaraproject/SimpleEdgeDiscovery/pull/66) specific to simple edge discovery service. we can also refer to other APIs like device status etc where the intents is to get information but POST and standard input body is use.``

@JoseMConde: I've read all the comments for Simple Edge Discovery and I've also read [API_Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#post-or-get-for-transferring-sensitive-or-complex-data) and if you read the @Kevsy [comment](https://github.com/camaraproject/EdgeCloud/discussions/330#discussioncomment-12005465) in the [discussion 330](https://github.com/camaraproject/EdgeCloud/discussions/330) you can see that we are following commonalities and the API Guidelines. It is true that the only API that has this behavior is Simple Edge Discovery and there is a trend to change these types of methods to Post, in which case we may have to change the API guidelines and make clear what behavior is expected from these types of methods.

> Another point is that Post/Edge-cloud-zones, doesn't have the same approach than Get/Edgecloudzones that we actually have in EAM. I think is interesting that the developer could filter by status. Also the ApplicationProfileId is a mandatory parameter, but maybe someone doesn't use Application Endpoint Registration and doesn't have this parameter or they use it for some Applications and for others they don't.
[Response from @maheshc01 ] i believe we discussed this during the call. based on what we agreed, Get/Edgecloudzones needs to be removed from AEM as it puts different intents related to orchestration and discovery into one yaml and as per our discussions we are separating it out to create yaml as per specific intents.
ApplicationProfileId is not from application end point registration. its will be from Application profiles API. the latest update on the application profile scope expansion is that its been discussed in api backlog group and is expect to be on the TSC agenda for approval soon. the parameter being mandatory, to discovery the optimal edge cloud zone we need additional information about the application and hence applicationProfileId is mandatory. If developers dont want to use this approach then they would use the simple edge discovery API>

@JoseMConde: We discussed this during the call, but I think I understood you wrong, sorry. Totally agree that we need to remove GET/edge-cloud-zones from EAM but my idea was that we agreed to add to Optimal Edge Discovery, what I saw is the behavior of the Get/edge-cloud-zones that know exists in OED has not the same approach than the one that we had in EAM, because the GET in OED is focus on get the Optimal Edge Cloud Zone after the instantiation of the Application and in EAM we need to know the available Edge Cloud Zones before the instantiation. So I think we have to discuss about how to manage this because I am affraid we will need to different methods.
Related to ApplicationProfileId, I meant Application Profiles,I got the name wrong, and I was thinking in the issue mentioned before that we need in EAM to know the available Edge Cloud Zones before the instantiation, for this I didn't see this parameter as mandatory.

> Analizing the behaviour of POST/app-endpoints in this API and the behaviour of Application Endpoint Discovery, could be almost the same, the only thing taht is different is that we use AppId to know wich is the Application that we need the endpoint and this API uses ApplicationEndpointsId, am I wright or I am missing something??
[Response from @maheshc01 ] as discussed during our meetings, this end point needs to be removed from this optimal edge discovery yaml and the related updated need to be done on AED yaml. we will need a different PR for this. @kohlian18 could you create a different PR for this.>

@JoseMConde: Totally agree with you, I can help you @kohlian18 with the PR and with the update of the YAML.

JoseMConde:
Hi @kohlian18 @maheshc01  , some ideas about this API:
- Trying to align with the current Get/Edge-cloud-zones that we have in EAM, I am wondering if make sense to have two different methods, because the approach is a little bit different in my view. The discovery we have in OED returns a list of optimal edge cloud zones where you can register your deployed application, but the discovery that we have in EAM returns the list of Edge Cloud Zones availables where the developer can deploy the application, so It could be a Get, because we are not going to use Device as a parameter. We filter by status and also by region because we would like to provide the developer to filter by those parameters. It is not the optimal based on the location of the UE or the Application profile, it is just the list of ECZ.
- Also in EAM we have Discover Cluster that maybe make sense to have also in this API and not in EAM.
image
What do you think??

gainsley:
Just some thoughts on the on-going discussion.

My interpretation of "discovery" is that the system is going to perform some logic based on some set of input data, and generate a set of matching options. For example, we want to discover the optimal edge-cloud-zone based on a device's location, and the target application's requirements (cpu/mem/gpu/latency/VM-vs-container/cost/etc). The system is going to filter out non-matching edge-cloud-zones, and return a list of matching edge-cloud-zones potentially with information about each one's latency, cost, etc, so the client can decide which zone to deploy to (maybe client prefers to minimize latency, or maybe they prefer to minimize cost, for example).

This is very different from the EAM APIs for GET/edge-cloud-zones and GET/clusters. In my opinion these are not discovery, this is just a data dump of the existing objects. There is no input from the user, and there is no logic being performed. A user can use these to see which zones are on the system, but there are no checks if a user's application can actually be deployed to any of them.

I think the discovery APIs should be something like POST/optimal-edge-cloud-zones or POST/discover-edge-cloud-zones.

One other question I have is how big an area do you guys consider a zone covering? A zone is already made up of multiple edge-cloud-sites that are hidden from the user, so when a user deploys to a zone, the system is already performing some logic to figure out which edge-cloud-site in the zone to deploy to. Are zones small enough that there will actually be a bunch of choices for a device? Or are they large enough that really most of the time there will only be one obvious choice, unless a device is near a boundary between two zones?

maheshc01:
Hi @gainsley , @JoseMConde  and others,

I went over the discussion related to having GET/edge-cloud-zones in EAM as well as  POST/optimal-edge-cloud-zones in Optimal Edge discovery service. I was hoping to avoid operators having to support multiple and different APIs all related to similar capability of return a list of edge cloud zones. As of now we have 3, simple edge discovery service, optimal edge discovery service and GET/edge-cloud-zones in EAM . While there are minor differences in the intent for each of the APIs there is still overlaps in what we are trying to address via these APIs. This will increase the burden on both Operators as well as API consumers.
Operators will have to evaluate and decide which are all the variations of the API they would like to implement and support. API consumers have to similarly have to understand which of the APIs they will have to use based on their use case.
I was hoping to reduce the number of variations by trying to combine the features. For example, by making the input fields in Simple discovery service, we could make it respond similar to GET/edge-cloud-zones in EAM  but the API would still need to be a POST call.
But if you all think having all of them as different APIs, i am good to go with the team's consensus on this.

maheshc01:
@JoseMConde regarding the megalinter not run on this and as pointed out by Fabrizio, i created this feature branch and was using this to coordinate fixes along woth @kohlian18 . I can go ahead and merge these changes to the feature branch and create a new PR for main branch.
let me know if you want me to do it now or after the discussion on GET/edge-cloud-zones in EAM  is completed.

maheshc01:
@JoseMConde regarding the megalinter not run on this and as pointed out by Fabrizio, i created this feature branch and was using this to coordinate fixes along with @kohlian18 . I can go ahead and merge these changes to the feature branch and create a new PR for main branch.
let me know if you want me to do it now or after the discussion on GET/edge-cloud-zones in EAM  is completed.

JoseMConde:
Hi @maheshc01 , I think we can solve the discussion about GET/edge-cloud-zones first and then we will update the PR.

JoseMConde:
Hi @camaraproject/edge-cloud_codeowners , as we commented in the last meeting we were going to use this PR to discuss about Get/edge-cloud-zones, beacuse we don't have MegaLinter.
I think everyine is happy keeping Get/edge-cloud-zones inside EAM, so my propose is to merge this PR and then @maheshc01 will open a new one attached to the main branch with MegaLinter.

Kevsy (approved):

PR #333, opened 2025-02-03
EAM API: fixes for app instance post (alternative to #316)

gainsley:
#### What type of PR is this?

* correction
* enhancement/feature
* documentation

- Clarifies the expected behavior when deploying multiple app instances via the appinstances POST API
- Generalizes the behavior of the API as a "bulk" API, which is just multiple of what would have been a singular API
- Assign a name to the AppInstance. Without this we cannot have declarative/infrastructure-as-code clients layered on top of the NBI API. See https://github.com/camaraproject/EdgeCloud/issues/255. Also makes it consistent with other objects that also have names, i.e. Apps, EdgeCloudZones.
- Adds more necessary information to the returned AppInstanceInfo, including the AppId (previously no way to figure out which App the AppInstance was an instance of). Makes the necessary information required. Changed the EdgeCloudZoneName to an EdgeCloudZoneId for consistency, as we always reference other objects by their ID, not their name.
- Adds a status code and message to each deploy response to allow for distinguishing the results of each deployment request separately.

Note that this is an alternative approach to #316 which attempted to address the same issues but using a singular API approach. This PR is the bulk API approach which more closely resembles the original intent of the API.

Fixes https://github.com/camaraproject/EdgeCloud/issues/256, https://github.com/camaraproject/EdgeCloud/issues/255

#### Special notes for reviewers:

One potential contentious change I have made is to include the AppId with each AppInstanceDeployRequest, rather than make it common across all requests. The primary reason was to generalize the handling of "bulk" APIs, such that a bulk API simply has request/response objects which are arrays of what the singular API would be. This approach could be extended to other APIs (although I don't really see any candidates at the moment). This also makes it more flexible, allowing a user to deploy different AppIds across various edge-sites in a single API call, given that the intent here is to avoid the overhead of making multiple API calls to deploy app instances.

#### Changelog input

```
 release-note
- clarify and fix behavior of appinstances POST

```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -370,49 +370,46 @@ paths:
         - Application
       summary: Instantiation of an Application
       description: |
-        Ask the Edge Cloud Platform to instantiate an application to one
-        or several Edge Cloud Zones with an Application as an input and an
-        Application Instance as the output.
-      operationId: createAppInstance
+        A bulk request consisting of one or more application deployment
+        requests to deploy an application to a target Edge Cloud Zone
+        or Cluster. If the overall request is accepted, each individual
+        request is handled independently, and the response will be an
+        array of deployment responses, each with their own individual
+        status codes. The array of deploy responses will be the same
+        size as the input request array, and location of each response
+        in the array will match the location of the corresponding
+        request in the input array.
+      operationId: createAppInstances
       parameters:
         - $ref: '#/components/parameters/x-correlator'
       requestBody:
         description: |
-          The Application ID and the array of Edge Cloud Zones to deploy
-          it to.
+          An array of deployment requests indicating where to
+          deploy an application.
         content:
           application/json:
             schema:
-              type: object
-              required:
-                - appId
-                - appZones
-              properties:
-                appId:
-                  $ref: '#/components/schemas/AppId'
-                appZones:
-                  $ref: '#/components/schemas/AppZones'
+              type: array
+              items:
+                $ref: '#/components/schemas/AppInstanceDeployRequest'
+              minItems: 1
         required: true
       responses:
         '202':
-          description: Application instantiation accepted
+          description: |
+            An array of responses corresponding to the input array of
+            requests. The location in the array of each response
+            corresponds to a request from the same location in the array
+            of requests.
           headers:
             x-correlator:
               $ref: "#/components/headers/x-correlator"
-            Location:
-              description: Contains the URI of the newly created application.
-              required: true
-              schema:
-                type: string
           content:
             application/json:
               schema:
-                type: object
-                properties:
-                  appInstances:
-                    type: array
-                    items:
-                      $ref: '#/components/schemas/AppInstanceInfo'
+                type: array
+                items:
+                  $ref: '#/components/schemas/AppInstanceDeployResponse'
                 minItems: 1
         '400':
           $ref: '#/components/responses/400'
@@ -420,17 +417,8 @@ paths:
           $ref: '#/components/responses/401'
         '403':
           $ref: '#/components/responses/403'
-        '409':
-          description: Conflict
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorInfo'
-              example:
-                status: 409
-                code: CONFLICT
-                message: "Application already instantiated in the given
-                  Edge Cloud Zone or Edge Cloud Region"
+        '404':
+          $ref: '#/components/responses/404'
         '500':
           $ref: '#/components/responses/500'
         '501':
@@ -724,6 +712,54 @@ components:
         Edge Cloud Platform generates this identifier when the
         Application is submitted.
 
+    AppInstanceDeployRequest:
+      description: |
+        A request to deploy an application instance of the specified
+        appId. Either the EdgeCloudZoneId or the KubernetesClusterRef
+        must be specified.
+      type: object
+      required:
+        - appId
+        - name
+      properties:
+        appId:
+          $ref: '#/components/schemas/AppId'
+        name:
+          $ref: '#/components/schemas/AppInstanceName'
+        edgeCloudZoneId:
+          $ref: '#/components/schemas/EdgeCloudZoneId'
+        kubernetesClusterRef:
+          $ref: '#/components/schemas/KubernetesClusterRef'
+
+    AppInstanceDeployResponse:
+      description: |
+        Response information for a given AppInstanceDeployRequest.
+        A 202 accepted code indicates a successful response, and the
+        appInstanceInfo must be set. All other response codes indicate
+        an error with the message providing detailed information,
+        and the appInstanceInfo may be omitted unless it has been
+        created and left in an error state.
+      type: object
+      required:
+        - code
+      properties:
+        code:
+          type: integer
+          description: |
+            HTTP status code describing the result of deploying
+            this instance. Code 202 indicates the deployment
+            is accepted and in progress. Code 400 indicates a
+            badly formed request. Code 409 indicates an
+            instance with the name already exists. Other codes
+            may be used to indicate some form of error.
+        message:
+          type: string
+          description: |
+            Message providing further details about the result
+            of the request.
+        appInstanceInfo:
+          $ref: '#/components/schemas/AppInstanceInfo'
+
     AppInstanceId:
       type: string
       format: uuid
@@ -736,9 +772,21 @@ components:
     AppInstanceInfo:
       description: Information about the application instance.
       type: object
+      required:
+        - name
+        - appId
+        - appInstanceId
+        - appProvider
+        - edgeCloudZoneId
       properties:
+        name:
+          $ref: '#/components/schemas/AppInstanceName'
+        appId:
+          $ref: '#/components/schemas/AppId'
         appInstanceId:
           $ref: '#/components/schemas/AppInstanceId'
+        appProvider:
+          $ref: '#/components/schemas/AppProvider'
         status:
           description: Status of the application instance (default is 'unknown')
           type: string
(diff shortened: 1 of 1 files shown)


JoseMConde:
@camaraproject/edge-cloud_codeowners Please, take a look to this PR and give some feedback.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -370,67 +370,55 @@ paths:
         - Application
       summary: Instantiation of an Application
       description: |
-        Ask the Edge Cloud Platform to instantiate an application to one
-        or several Edge Cloud Zones with an Application as an input and an
-        Application Instance as the output.
-      operationId: createAppInstance
+        A bulk request consisting of one or more application deployment
+        requests to deploy an application to a target Edge Cloud Zone
+        or Cluster. If the overall request is accepted, each individual
```
gunjald, line 375: I think we should qualify term "Cluster" e.g. "Set of one or more Edge Cloud Zones" as formally we have not introduced the term cluster for any of the APIs

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -370,67 +370,55 @@ paths:
...
+        request is handled independently, and the response will be an
+        array of deployment responses, each with their own individual
+        status codes. The array of deploy responses will be the same
+        size as the input request array, and location of each response
+        in the array will match the location of the corresponding
```
gunjald, line 380: I think the condition "location of each response in the array will match the location of the corresponding request" may be relaxed by introducing for example unique identifier for each request so the individual status of each request can be correlated with the individual response. This is just a suggestion to consider as such strict ordering at times create complexity at the application-level.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -724,6 +712,54 @@ components:
         Edge Cloud Platform generates this identifier when the
         Application is submitted.
 
+    AppInstanceDeployRequest:
+      description: |
+        A request to deploy an application instance of the specified
+        appId. Either the EdgeCloudZoneId or the KubernetesClusterRef
+        must be specified.
+      type: object
+      required:
```
gunjald, line 721: Based on the description one of EdgeCloudZoneId or KubernetesClusterRef shall be present in the required list of parameters in addition to appId and name.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -724,6 +712,54 @@ components:
...
+        edgeCloudZoneId:
+          $ref: '#/components/schemas/EdgeCloudZoneId'
+        kubernetesClusterRef:
+          $ref: '#/components/schemas/KubernetesClusterRef'
+
+    AppInstanceDeployResponse:
+      description: |
+        Response information for a given AppInstanceDeployRequest.
+        A 202 accepted code indicates a successful response, and the
+        appInstanceInfo must be set. All other response codes indicate
```
gunjald, line 738: As the 202 status code indicate that the request is accepted but it doesnt imply that the app instantiation shall be tried immediately during the request handling. In that case presence of appInstanceInfo may not be a MUST condition as described. And there may be GET method available which may allow API consumer to retrieve the status of each unique request object at a later point of time where a successful instantiation of an app instance can be informed by including appInstanceInfo. Any thoughts?

code/API_definitions/Edge-Application-Management.yaml
gainsley: A cluster is a subset of resources within an Edge Cloud Zone, typically a Kubernetes cluster. I agree it should be well defined somewhere.

code/API_definitions/Edge-Application-Management.yaml
gainsley: I would argue the opposite, that it's easier to manage strict ordering, otherwise the code will end up with O(n^2) algorithms (nested for loops) in order to match requests to responses.

code/API_definitions/Edge-Application-Management.yaml
gainsley: If the status is 202, the appInstanceInfo is a MUST because it contains the appInstanceId. Without the appInstanceId there is no way for a follow-up GET to query the status.

code/API_definitions/Edge-Application-Management.yaml
gunjald: Actually my query was more towards the simplicity of handling every objects independently rather than for the performance per se. If as API invoker I am associating a objectID then from response i loop through the objects, pickup the objectId and retrieve local context. If the number of edge cloud zones are too large then there might the performance issue as you said but for a smaller number i think even a client side objectId is one of the option. But again ii is just an attempt to seek the best practices available for scenarios like this for object aggregation.

code/API_definitions/Edge-Application-Management.yaml
Kevsy: > If the status is 202, the appInstanceInfo is a MUST because it contains the appInstanceId. Without the appInstanceId there is no way for a follow-up GET to query the status.

`appInstanceId` applies to a single app instance. So the presence of each `appInstanceId`  in the response array implies that the request for that particular deployment on a given zone was successful.

But here you can have a situation where one ore more of the deployment requests fails, hence there is no app instance for those zones.

So for the response of `202`  the response body can return  `appInstanceId` for successful deployments and omit that for failed deployments. 

However - you may prefer to refactor this operation as an asynchronous request with callback.. That depends on how long the caller will wait before apps are deployed and app instance Ids created - which depends on the cloud provider, the number of edge cloud zones in the request etc.

gainsley: There might be some confusion here about which status code we're talking about. There's the status code for the overall HTTP request. Then for each AppInstance request, there is an additional status code for that specific AppInstance. Logically this API is really a batch API with multiple AppInstance deploy requests bundled together. That's part of the reason I somewhat prefer the other approach (#316) as it doesn't have this confusion.

The overall HTTP request can have a status code of 202. And each AppInstance request in the response can have it's own individual status code (202, 403, 409, etc). The individual status code replicates what would be the status code for a singular API call (like #316). My comment was for the individual status code, if it is 202, that means the individual request was accepted and the system is working on deploying that particular application instance, and the system must return an appInstanceId.

We could omit the individual status code and infer success/failure based on the presence/absence of the appInstanceId, but that:
- is less clear, as we're inferring one state from another field
- reduces the individual status code to a binary success/fail, rather than allowing for indicating other errors like permission denied, bad status request, etc
- deviates from the behavior of a single, non-batched API call, in case we ever decide to add one in addition to the batched api

Note that the overall status code is independent of the individual status codes. Some reasons that the overall status could fail:
- user does not have permission to the api
- JSON request body could not be parsed
Some reasons the individual appInstance requests could fail:
- targeted appId/cluster/edge-site does not exist
- user does not have permission to deploy appId or deploy to specified cluster/edge-site
- targeted cluster/edge-site does not have enough resources available, or user has reached their resource quota limit

Regarding callbacks, that can be an alternate approach to polling, but unfortunately callbacks won't work if the client is behind a NAT/firewall, which is often the the case when the user is a human, and the client is a command line app or web app. Callbacks are more appropriate when we expect the client to be another service that can have a listening port open and always reachable. We could in fact support both, but I think the intent here is that the user is typically a human, so I don't think it makes sense to add callback support when they most likely can't be used.

Kevsy:
Thanks @gainsley , that's helpful.

> There might be some confusion here about which status code we're talking about. 

My reply above was in the context of a single request to the CAMARA endpoint, which in turn triggered individual requests to deploy app instances (likely calling other, non-CAMARA APIs to do that). The initial request would return `202 Accepted` as long as it was valid/authenticated/authorised, even if all of the batched requests therein failed.

> Logically this API is really a batch API with multiple AppInstance deploy requests bundled together.

Agreed.  Of note is that the bundle of requests is likely to involve another system (i.e. if the edge provider is not the operator).  Hence, either there's a risk of long delay in returning an aggregated response which includes the outcome of each of the requests, or of returning an initial response quickly but which simply indicates the batch request has been accepted. 

> Then for each AppInstance request, there is an additional status code for that specific AppInstance. 

Agreed, looks like the `code` and `status` enumeration cover that.

My point was rather to do with `appInstanceId` for failed or pending requests, because the `Edge Cloud Provider generates this identifier.` So to include all the `appInstanceId`s in the overall response you would need to wait until all the individual requests have completed - which may not be a good developer experience.

> Regarding callbacks, that can be an alternate approach to polling, but unfortunately callbacks won't work if the client is behind a NAT/firewall, which is often the the case when the user is a human, and the client is a command line app or web app. Callbacks are more appropriate when we expect the client to be another service that can have a listening port open and always reachable. We could in fact support both, but I think the intent here is that the user is typically a human, so I don't think it makes sense to add callback support when they most likely can't be used.

Does Commonalities cloud events not account for that..?

seralogar:
I believe that to resolve this discussion we should explore a possible third way.

Instead of directly creating instances (one or many) and forcing consumers to implement status tracking/reacting logic, we could introduce a new object called deployment and let the operator take responsibility of instance creation. The operator knows the real status of the infrastructure and can act without consumer intervention. There's no need for the consumer to implement all the logic that surrounds the deployment of multiple instances of an app.

The response to the creation could be synchronous, as in, the deployment creation request has been received. In a successful deployment creation, the app instances will be created and linked to the deployment by the orchestrator—similarly to other cloud-native scenarios. Then, we could have endpoints for retrieving information about the deployment and its instances to get the specific app instance details and its statuses. This would serve as a preliminary way for the consumer to monitor the status of the instances—a rudimentary watcher that could evolve in the future (but that's another discussion).

Note that the new deployment definition is compatible with varying levels of sophistication in the operator's orchestrator. In its simplest form, it might merely create static instances, fulfilling both ideas of single or bulk instance creation mentioned in this PR; in a more advanced setup, it could support self-healing, migration, scaling, smart placement, etc. This approach not only simplifies client-side interactions but also provides a flexible mechanism to handle the potential complexities of app instance deployment.

@JoseMConde Perhaps it's interesting to discuss this possible approach in the bi weekly meeting with everyone.

JoseMConde:
Thanks @seralogar, I think is a very good idea, tomorrow I will reserve some slot in the meeting to discuss this topic.

seralogar:
That's a good point, thank you for the feedback.

I don't think it should be a requirement and we could keep the creation of instances without a Deployment (just like Pods, as you said). In that case, the /appinstance endpoint should become singular (I believe this is the intention of #316)

However, I still think it should be bundled as a new alternative PR, since the condition to redefine the /appinstance method as a singular endpoint is to keep the "bulk" functionality with a new object called Deployment. Otherwise, the merged version of this API would be an inconsistent state that redefines the intention of this API (perhaps this is just a formality and it can be done in multiple PRs, I am not sure if this is OK or not)

What do you think?

gainsley:
I do agree that a new "deployment" based feature should be it's own PR. Besides I think that starts to go beyond just implementing the bulk version of the singular appinstance create API, as it's adding a new data structure and concept to the API. I could see then adding more features/behavior tied to deployments, i.e. scaling, managing which zones are part of the deployment, etc.

However, I don't see any need to hold off on #316 or even #333 for that, as that state of current create app instance API is not really usable in my opinion, and should be made functional before we start adding another feature on top.

Something else to consider: a deployment may not strictly be equivalent to a bulk app instance create. For example, the bulk app instance create may allow different parameters per instance, i.e. different zones may have different GPU offerings/availability, so a user may wish to deploy each app instance with different resource configurations. I don't know if you are considering that as something a deployment can provide.

seralogar:
Sounds good. I propose then merging #316 and then I could create a new PR with Deployments built on top of a singular instance endpoint.

PS. About different parameters, in my view, different requirements equal different deployments, although in the future there could be an affinity mechanism to tolerate different requirements.

JoseMConde:
Hi @gainsley, It makes sense to me, we can approbe this PR and close #333. With the idea of "Deployment" the single instantiation could be the best option. WDYT??

gainsley:
@JoseMConde Sure, I'm fine with merging #316, closing #333, and the idea of a "deployment" approach for bulk instantiation, if that's the consensus. Sounds good to me. Thanks!

PR #334, opened 2025-02-05
Application Endpoints Discovery (AED) enhancements

kohlian18:
#### What type of PR is this?

* enhancement/feature

Changed the method of the API from GET /apps-endpoints to POST /application-endpoints with region as a filter. Added device information to the json request body. 

This helps Application endpoint discovery functionality to align with community API guidelines.


code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+  description: |
+    # Connect to the Optimal Application Endpoint
+    ---
+    # Overview
+    The Application Endpoint Discovery API extends beyond the capabilities of the
+    Simple Edge Discovery API by not only locating the nearest Edge Cloud Zone
+    but also directly linking to the application endpoints within those
+    Edge Cloud Zones.
+    This API intent enables immediate connection to the most relevant
+    service endpoint, optimizing the application's interaction with
```
gunjald, line 15: I think we should change the term "service endpoint" as it may not have defined earlier or in this file yet.

gunjald, line 15: "optimizing the application's" does it mean "optimizing the external application's"? 
Intent is to explicitly refer the an application component in terms of edge and outside edge. But please check for explicit usage of terms.

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    edge computing resources.
+    # Introduction
+    Operators often host multiple Edge Cloud Zones across various territories.
+    Connecting an application directly to the optimal Edge Cloud Zone
+    endpoint may minimize latency.
+    Unlike the Simple Edge Discovery API, which only provides the zone location,
+    this API provides the actual endpoints within the nearest
+    zone. This approach ensures that developers can instantly integrate these
```
gunjald, line 23: should we change zone to Edge Cloud Zone for consistency?

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    endpoints into their applications, improving the user
+    experience.
+    After identifying the nearest application endpoint through the API,
+    developers can:
+    * Connect their application client directly to the chosen application
```
gunjald, line 28: should we change application to edge application for consistency and clarity of terms?

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    endpoint, enhancing the responsiveness of the service.
```
gunjald, line 29: I think the term service is not yet defined. So please check what should be the appropriate term we want to use here.

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    * If no suitable endpoints are available, the
```
gunjald, line 30: Is the intent "If no suitable endpoints are available" is "If no suitable endpoints are available for the edge application in requested Edge Cloud Zone"... Just a suggestion to rephrase to be more explicit.

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    developer can use the API EdgeApplicationManagement in order to
+    instantiate a new instance.
+    Edge Cloud Zone, optimizing their service distribution based on real-time
+    network geography.
+    # Quick Start
+    This API can be utilized by both client applications on user devices and
+    by servers. It uses a straightforward endpoint:
+    `/application-endpoints`.
+    For accurate service localization, the API requires device-specific
```
gunjald, line 39: Please check the term service. If needed please use existing terms or define the service.

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    identifiers such as:
+    * `IP-Address`: Public IP address of the user device, which can be fetched
+    via
+    external IP discovery services.
+    * `Phone-Number` : The device's phone number in international E.164 format.
+    * `Network-Access-Identifier`: A unique identifier provided by the network
```
gunjald, line 45: As Phone-Number is defined above E.164 format do we also mention the format for Network-Access-Identifier in similar manner?

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+    
+    # Authorization and authentication
+    In order to maintain the integrity and security of the Application Discovery
+    API, access is governed by robust authentication and authorization
+    processes.
+    This ensures that only valid, authorized requests are able to interact with
+    the API and retrieve application endpoint information.
+    ## OpenID Connect
+    The API utilizes OpenID Connect on top of OAuth 2.0 for user authentication.
+    Users must register to obtain client credentials, which are necessary to
```
gunjald, line 67: In context of oAuth2.0 i suppose we should mention who the term "Users" refers to. Is it for external applications which needs to obtain client credentials from operator?

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+            the same value that was used in the request. Otherwise, it is
+            optional to include the "x-correlator" header in the response with
+            any valid value. Recommendation is to use UUID for values.
+          schema:
+            type: string
+        - name: Region
+          in: query
+          required: false
+          description: |
+            The common name of the closest Edge Cloud Zone to the user device.
```
gunjald, line 130: Seems the description suggesting Region equivalent to Edge Cloud Zone. If that is the case it should be mention here. Otherwise if the two identifiers are different then description should state it.

code/API_definitions/Application_endpoint_discovery.yaml
```diff
@@ -0,0 +1,685 @@
...
+        When the API Consumer includes the "x-correlator" header in the request,
+        the API provider must include it in the response with the same value t
+        hat was used in the request. Otherwise, it is optional to include the
+        "x-correlator" header in the response with any valid value.
+        Recommendation is to use UUID for values.
+      required: false
+      schema:
+       type: string
+  schemas:
+    EdgeHostedapplication:
```
gunjald, line 244: Should the application be capitalized as in EdgeHostedApplication?

code/API_definitions/Application_endpoint_discovery.yaml
kohlian18: okay, i will change that

code/API_definitions/Application_endpoint_discovery.yaml
kohlian18: sure, i will do that

PR #335, opened 2025-02-05
AED enhancements

kohlian18:
What type of PR is this?

enhancement/feature
Changed the method of the API from GET /apps-endpoints to POST /application-endpoints with region as a filter. Added device information to the json request body.

What this PR does / why we need it:
This helps Application endpoint discovery functionality to align with community API guidelines.


JoseMConde:
Hi @kohlian18, some comments:
- You changed directly the YAML, so It's quite difficult to analyze the changes.
- The goal of our view of AED is:
image
 So we need to be able to filter for the closest Edge Cloud Zone, we don't need the filter by region, this is needed in OED, there(OED) we need to know what Edge Cloud Zones we have in a determinated region.
- Also we need to add the parameter AppId, because in our view, we don't use Application Endpoint registration, so we don't have applicationEndpointsId but we identify the App with the AppId.
- With the same vision, we also think that applicationEndpointsId should be optional.
- My idea is that with this API can return the closest App Instance Endpoint Zone to the user device as It does now, and also returns a list of optimal Application Endpoints that client devices can connect to.

JoseMConde:
Also I see Nicola still as a reviewer, he is not longer part of this project, @seralogar is now codeowner and I don't see him.
Could be some issue with the PR?? Becasuse I think the others PRs recently open are OK, anyone knows what could be the problem??

gainsley:
Because the file was renamed and there were a lot of changes, github thinks it's a completely new file instead of a modification of the existing file. To get around this, I would suggest having separate PRs for the file name change vs the file modifications, that way github won't get confused will give you a proper diff.

kohlian18:
> Because the file was renamed and there were a lot of changes, github thinks it's a completely new file instead of a modification of the existing file. To get around this, I would suggest having separate PRs for the file name change vs the file modifications, that way github won't get confused will give you a proper diff.

i didnt make any changes directly to the file but i can close this PR and raise a new one if that helps

kohlian18:
> Also I see Nicola still as a reviewer, he is not longer part of this project, @seralogar is now codeowner and I don't see him. Could be some issue with the PR?? Becasuse I think the others PRs recently open are OK, anyone knows what could be the problem??

Yes there seems to be an issue with the PR, can you please help me close it?

PR #346, opened 2025-04-03
Linting config update

rartych:
#### What type of PR is this?

Add one of the following kinds:
* subproject management

#### What this PR does / why we need it:

.spectral.yml is outdated
.yamllint.yaml is not present, hence the default config is used for yamllint 

#### Special notes for reviewers:
 The problem was discovered when #337 and #321 were analysed

#### Changelog input

```
Megalinter config files updated
```

PR #347, opened 2025-04-04
Updated the number of APIs in Application-endpoint-registration.yaml

urvika-v:
#### What type of PR is this?

* correction

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/Application-endpoint-registration.yaml
+++ b/code/API_definitions/Application-endpoint-registration.yaml
@@ -28,7 +28,7 @@ info:
 
     # API functionality
 
-    The API provides 4 operations:
+    The API provides 5 operations:
     - POST registerApplicationEndpoints: Registers the Application Endpoints
                     of a deployed application to a specified edge cloud zone.
     - GET getAllRegisteredApplicationEndpoints: Returns endpoint information for


code/API_definitions/Application-endpoint-registration.yaml
```diff
@@ -28,7 +28,7 @@ info:
 
     # API functionality
 
-    The API provides 4 operations:
+    The API provides 5 operations:
     - POST registerApplicationEndpoints: Registers the Application Endpoints
```
gunjald, line 32: change look fine but i have one suggestion. Do we really need to put a number for the supported operations or we can just say the API provides following operations? So in future if we add more operations we do not need to track the count. Just a suggestion though.

urvika-v: Thank you for the suggestion. Modified it.

gunjald (approved):
Seems fine to me.

maheshc01:
@JoseMConde please avoid merging code which hasnt gone through linting check. we have to first merge PR #346 to fix the linting issue and then we can push in new commits.

PR #356, opened 2025-04-29
EAM API: Add deployment entity for bulk edge cloud zone instantiation.

seralogar:
#### What type of PR is this?
* enhancement/feature

Introduce Deployment resource for multi-zone instance management

As discussed in https://github.com/camaraproject/EdgeCloud/pull/333#issuecomment-2729012565, this PR adds a new `/deployments` endpoint to the Edge Application Management API.

This introduces a `Deployment` entity that allows users to specify requirements for deploying and managing application instances across multiple edge zones. The user can use this specification to select which set of edge cloud zones should have an instance of a particular application running, and the system should be able to manage the bulk creation and its lifecycle.

The existing functionality for creating single instances directly via the `/instances` endpoint remains unchanged and, in fact, could be leveraged by the system to enforce the specification stated in the deployment resource.

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/Edge-Application-Management.yaml
+++ b/code/API_definitions/Edge-Application-Management.yaml
@@ -744,10 +744,10 @@ paths:
         Update the configuration or properties of an existing application deployment
         using JSON Merge Patch semantics (RFC 7396). Only the fields provided in the
         request body will be updated. Fields not included in the request will remain unchanged.
-        
+
         IMPORTANT: When updating array fields (like edgeCloudZones or kubernetesClusterRefs),
         JSON Merge Patch will REPLACE the entire array, not merge or append to it.
-        
+
         This operation may include changing the deployment name, target Edge Cloud Zones, or other updatable fields.
       operationId: updateAppDeployment
       parameters:
@@ -763,7 +763,7 @@ paths:
         description: |
           The fields to update for the application deployment using JSON Merge Patch (RFC 7396).
           Only the fields included in the request will be updated; omitted fields remain unchanged.
-          
+
           NOTE: When updating array fields (edgeCloudZones, kubernetesClusterRefs), the entire array
           will be REPLACED, not merged. To modify an array, you must include the complete array
           with all desired elements in your request.
@@ -790,7 +790,7 @@ paths:
                   This example shows how to update only the deployment name.
                   Other fields will remain unchanged.
                 value:
-                  appDeploymentName: "my-updated-deployment"
+                  appDeploymentName: "my_updated_deployment"
               updateMultipleFields:
                 summary: Update multiple fields simultaneously
                 description: |
@@ -798,34 +798,34 @@ paths:
                   and Edge Cloud Zones in a single request. Remember that both
                   array fields will be completely replaced with the new values.
                 value:
-                  appDeploymentName: "production-deployment"
+                  appDeploymentName: "production_deployment"
                   edgeCloudZones:
-                    - "zone-us-east-1"
-                    - "zone-us-west-1"
+                    - "123e4567-e89b-12d3-a456-426614174000"
+                    - "123e4567-e89b-12d3-a456-426614174001"
                   kubernetesClusterRefs:
-                    - "cluster-prod-east"
-                    - "cluster-prod-west"
+                    - "642f6105-7015-4af1-a4d1-e1ecb8437abc"
+                    - "642f6105-7015-4af1-a4d1-e1ecb8437def"
               arrayReplacementExample:
                 summary: Example of array replacement behavior
                 description: |
                   This example demonstrates how arrays are completely replaced in JSON Merge Patch.
-                  
+
                   If the current deployment has:
-                  - edgeCloudZones: ["zone-us-east-1", "zone-us-west-1", "zone-eu-central-1"]
-                  - kubernetesClusterRefs: ["cluster-1", "cluster-2"]
-                  
+                  - edgeCloudZones: ["123e4567-e89b-12d3-a456-426614174000", "123e4567-e89b-12d3-a456-426614174001", "123e4567-e89b-12d3-a456-426614174002"]
+                  - kubernetesClusterRefs: ["642f6105-7015-4af1-a4d1-e1ecb8437abc", "642f6105-7015-4af1-a4d1-e1ecb8437def"]
+
                   And the user sends this patch:
-                  - edgeCloudZones: ["zone-us-east-1", "zone-ap-south-1"]
-                  
+                  - edgeCloudZones: ["123e4567-e89b-12d3-a456-426614174000", "123e4567-e89b-12d3-a456-426614174003"]
+
                   The result will be:
-                  - edgeCloudZones: ["zone-us-east-1", "zone-ap-south-1"] (completely replaced)
-                  - kubernetesClusterRefs: ["cluster-1", "cluster-2"] (unchanged, as it wasn't in the patch)
-                  
+                  - edgeCloudZones: ["123e4567-e89b-12d3-a456-426614174000", "123e4567-e89b-12d3-a456-426614174003"] (completely replaced)
+                  - kubernetesClusterRefs: ["642f6105-7015-4af1-a4d1-e1ecb8437abc", "642f6105-7015-4af1-a4d1-e1ecb8437def"] (unchanged, as it wasn't in the patch)
+
                   To add a single zone while keeping existing ones, ALL zones must be included in the request.
                 value:
                   edgeCloudZones:
-                    - "zone-us-east-1"
-                    - "zone-ap-south-1"
+                    - "123e4567-e89b-12d3-a456-426614174000"
+                    - "123e4567-e89b-12d3-a456-426614174003"
       responses:
         '200':
           description: Application deployment updated successfully


JoseMConde:
LGTM

JoseMConde (approved):

albertoec (commented):
Hello @seralogar !
Thanks for the contribution. it makes perfect sense to me. I just left a couple of minor comments for you to check and see if they make sense from your side.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,173 @@ paths:
...
+            - edge-application-management:deployments:read
+      tags:
+        - Application
+      summary: Retrieve a list of Application Deployment for a given App
+      description: |
+        Ask the Edge Cloud Provider the information of the Deployments for a
+        given application
+      operationId: getAppDeployments
+      parameters:
+        - $ref: '#/components/parameters/x-correlator'
```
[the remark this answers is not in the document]
gainsley: I agree, I think it's a good idea to be able to query by deployment name or deployment id.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,173 @@ paths:
...
+            Identificator of the specific application deployment
+            that will be terminated
+          required: true
+          schema:
+            $ref: "#/components/schemas/AppDeploymentId"
+      responses:
+        '204':
+          description: |
+            Request accepted to be processed. It applies for async
+            deletion process
```
gainsley, line 709: I think this should be 202, not 204. 204 indicates there's no further action needed, but in this case I believe we are assuming the API call will return immediately and the back-end will process the delete asynchronously. Thus to determine if the delete actually succeeded or not, the client will need to poll the GET API.

gainsley (commented):
Thanks for adding this!
- I see that a deployment is just a list of AppInstances, which makes sense. When I delete a deployment, does it delete all associated AppInstances, or do they remain and can be managed (deleted/updated/etc) individually?
- What happens if I delete an AppInstance from a deployment directly via `DELETE instances/appInstanceID`? Is that allowed? If allowed, does the deployment retain the AppInstance ID for it, or does it clean up the zone/appInstanceID from the deployment?
- Is there any plan to add an Update API, i.e. to change the zones for the deployment? I.e. if I want to deploy the same app to another zone, do I need to create a new deployment, or can I update my existing deployment? If I want to reduce the number of zones, I need some way to do that without deleting the instances in the deployment that I want to keep.

As you can see from my comments above, a simple bulk API is nice because there is no persistent data associated with it - it just creates a bunch of instances, and then you are free to manage those instances individually. The deployment approach, however, has a persistent data structure, so typically you would manage the instances via the deployment, rather than directly (otherwise it can get complicated for both the user and the implementing logic to manage two different ways of managing the same objects).

code/API_definitions/Edge-Application-Management.yaml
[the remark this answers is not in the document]
seralogar: Yes, I will add it.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -74,6 +74,14 @@ info:
...
     Cloud Zone.
 
+    __Application Deployment Management__
+    * __createAppDeployment__ - Requests the Edge Cloud Provider to create and
+    maintain application instances to multiple Edge Cloud Zones.
+    * __getAppDeployments__ - Retrieves a list of deployments for a given
+    application.
+    * __deleteAppDeployment__ - Terminates a specific application deployment,
+    removing all associated instances within the specified Edge Cloud Zone.
+
```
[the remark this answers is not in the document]
seralogar: True. I will change it.

code/API_definitions/Edge-Application-Management.yaml
seralogar: Yes, makes sense. I will change it.

seralogar:
Thank you @albertoec and @gainsley for your comments.

My view on @gainsley review comments: 

> * I see that a deployment is just a list of AppInstances, which makes sense. When I delete a deployment, does it delete all associated AppInstances, or do they remain and can be managed (deleted/updated/etc) individually?
> * What happens if I delete an AppInstance from a deployment directly via `DELETE instances/appInstanceID`? Is that allowed? If allowed, does the deployment retain the AppInstance ID for it, or does it clean up the zone/appInstanceID from the deployment?

I think the manager/owner/controller of those AppInstances should be the Deployment (a la kubernetes), and not the user, hence a DELETE of the owner makes a cascade deletion of its child resources (at least in the basic scenario). 

If someone removes an AppInstance that is a child resource of a Deployment, or it is lost for some reason, the Deployment is responsible to recreate it and keep it running (again, a la kubernetes). 

> * Is there any plan to add an Update API, i.e. to change the zones for the deployment? I.e. if I want to deploy the same app to another zone, do I need to create a new deployment, or can I update my existing deployment? If I want to reduce the number of zones, I need some way to do that without deleting the instances in the deployment that I want to keep.

This is a very interesting point. I didn't initially add an Update endpoint since the rest of resources didn't have one, but this makes a lot of sense and in fact, as your example states, is quite critical. I believe I can add it to this PR with the rest of the changes.

> As you can see from my comments above, a simple bulk API is nice because there is no persistent data associated with it - it just creates a bunch of instances, and then you are free to manage those instances individually. The deployment approach, however, has a persistent data structure, so typically you would manage the instances via the deployment, rather than directly (otherwise it can get complicated for both the user and the implementing logic to manage two different ways of managing the same objects).

It is true that a simple bulk API has a simpler logic for the orchestrator but it doesn't help the user in managing them. A Deployment opens the door for requirements/policy based orchestration and reduces the burden of managing single instances and keeping track of its statuses, something that in my view is critical and also the basis of any (edge)-cloud native system, while also providing bulk instantiation. It is true that what we're proposing here might be more complex in terms of implementation but I think the user can definitely benefit from it.

Thanks again for your comments, I will make the changes requested ASAP.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,173 @@ paths:
...
           $ref: '#/components/responses/503'
+
+  /deployments:
+    post:
+      security:
+        - openId:
+            - edge-application-management:deployments:write
+      tags:
+        - Application
+      summary: Deployment of an Application
```
Kevsy, line 567:
```suggestion
      summary: Deploy an Application
```

Kevsy: Changed from 'deployment' to make explicit this is an action, rather than information about a deployment.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,173 @@ paths:
...
+              $ref: "#/components/headers/x-correlator"
+        '400':
+          $ref: '#/components/responses/400'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+        '500':
```
Kevsy, line 721: If my comment above is accepted we can add 410 Gone here.

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,173 @@ paths:
...
+  /deployments/{appDeploymentId}:
+    delete:
+      security:
+        - openId:
+            - edge-application-management:deployments:delete
+      tags:
+        - Application
+      summary: Terminate an Application Deployment
+      description: |
+        Delete a deployment terminating all related instances of an application within
```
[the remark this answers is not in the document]
seralogar: It should remove `all related instances of an application`. I'll delete the rest.

code/API_definitions/Edge-Application-Management.yaml
seralogar: I will add it. I think it makes sense to add it to the responses to all GET,UPDATE,DELETE.

code/API_definitions/Edge-Application-Management.yaml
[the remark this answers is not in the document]
seralogar: I think we can keep the querystring for consistency with `/instances` for this PR, and we can discuss the `/{appId}/` approach in a different one, since perhaps it makes sense to change it for `/instances` as well?

JoseMConde:
I think everything is solved, I can't review the PR, you dismissed my review @seralogar, to be able to review again you should ask me again :) , I think we can approve and merge this PR.

JoseMConde (approved):

code/API_definitions/Edge-Application-Management.yaml
```diff
@@ -543,6 +556,257 @@ paths:
...
+          $ref: '#/components/responses/503'
+    patch:
+      security:
+        - openId:
+            - edge-application-management:deployments:update
+      tags:
+        - Application
+      summary: Update an Application Deployment
+      description: |
+        Update the configuration or properties of an existing application deployment.
```
gunjald, line 744: One small observation on explaining "changing the deployment name, target Edge Cloud Zones". In case of target Edge Cloud Zones when the array of EdgeCloudZoneId is present, does it mean it will have an additive effect that the given zone will be considered as additional zone to create new application deployment on these zone? In that case I suppose "changing target edge zones" may be equivalent to "add more edge cloud zones" but there may be no deletion of edge cloud zone in case of modification. Please check if we need to update any description for better explaining the given operations for this method.

seralogar: Ah, I see the problem now. PATCH is indeed allowing partial update for the resource, and in our case this would still be the case since you're not required to provide the whole object. However JSON Patching allows for different interpretations:
 
I am proposing a simpler approach, in line with [JSON Merge Patch](https://www.rfc-editor.org/rfc/rfc7396.txt), where arrays get replaced. 
I understand that you have in mind something more in line with [JSON Patch](https://www.rfc-editor.org/rfc/rfc6902), which requires defining the operations for different resources.
Kubernetes have their own patch implementation, and they support 4 different strategies, including the above.

I think this raises the question: which one should we support? 

I can't find any other example in CAMARA. I only found some [QoD](https://github.com/camaraproject/QualityOnDemand_PI1/blob/832f9ed4e0e8b94b03eb0dbb12c694e16e064754/3gpp/src/main/resources/TS29514_Npcf_PolicyAuthorization.yaml#L312) resources that support JSON Merge Patch.

I believe JSON Merge Patch is simpler and covers most of our use-cases, but probably more limited than JSON Patch.

Any opinion on this? Thanks for bringing this up, I completely missed it in my specification.

gunjald: Do we need to include an example to show how to use the parameters? Just a documentation suggestion.

seralogar: Yes! Good call. I have improved descriptions and added examples [here](https://github.com/camaraproject/EdgeCloud/pull/356/commits/ab0d29461a2c9d23e8056f365f90ed514015bb6d).

gunjald (approved):
To me changes looks fine.

JoseMConde:
Finally, I think we can approve and merge this PR. :) 
Any other comment @camaraproject/edge-cloud_codeowners ??

PR #358, opened 2025-05-09
Application Endpoint Discovery Test definitions

JoseMConde:
Application Endpoint Discovery Test definitions

#### What type of PR is this?

* documentation

Adding trhe file of Test Definitions.

```
docs

```


Kevsy (approved):
Approved with some changes made:
- style changed following lint results from .gherkin-lintrc
- three step scenarios had the same name, so have given them distinct names

